PR #1, opened 2022-12-16
Move cloud src code

yinweisu:
* Move cloud src code from `autogluon` to `autogluon-cloud`
* This PR move the code unchanged. Any changes needed will be sent in follow-up PRs.
* CI code not moved yet as it will involve changes too


setup.py
```diff
@@ -0,0 +1,52 @@
...
+spec.loader.exec_module(ag)
+###########################
+
+version = "0.1"
+version = ag.update_version(version, use_file_if_exists=False, create_file=True)
+
+submodule = "cloud"
+install_requires = [
+    # version ranges added in ag.get_dependency_version_ranges()
+    "autogluon.common<0.7",
```
Innixma, line 23: Are you sure we should have an open version range for the lower end here?

setup.py
```diff
@@ -0,0 +1,52 @@
...
+    "boto3",
+    "numpy",
+    "packaging<22",
```
Innixma, line 26: Are you sure we should have an open version range for the lower end here?

setup.py
```diff
@@ -0,0 +1,52 @@
...
+    "pandas",
+    # TODO: update to the latest after 0.6 container is out
+    "sagemaker>=2.94",
+    "pyarrow>=9.0",
+    "Pillow>=9.3.0",
```
Innixma, line 31: Maybe cap to some kind of future major version?

Innixma (approved):
LGTM!

Couple things that would be good to clarify:
1. What is different from the code in autogluon, if anything
2. How will the unit testing / CI work? (Can be added later too)

Finally, the review comments can be addressed in a follow-up PR to avoid altering the alignment between the code in AG and here.

yinweisu:
@Innixma Will be addressing comments in follow-up PRs. Thanks!

setup.cfg
```diff
@@ -0,0 +1,88 @@
+[pycodestyle]
+max_line_length = 160
```
tonyhoo, line 2: I assume these are consistent with what we have in AG?

setup.py
tonyhoo: +1, we shouldn't cap this unless there is a strong reason for it

src/autogluon/__init__.py
```diff
@@ -0,0 +1 @@
+__import__("pkg_resources").declare_namespace(__name__)
```
tonyhoo, line 1: Shall we change the package name from `autogluon` to `autogluon-cloud`?

src/autogluon/cloud/__init__.py
```diff
@@ -0,0 +1,5 @@
+from autogluon.common.utils.log_utils import _add_stream_handler
```
tonyhoo, line 1: Let's remove the `cloud` sub-package as it is a standalone repo now

setup.py
yinweisu: OK. I think we need a lower end here: `>=0.6`, but probably not a upper cap

setup.cfg
Innixma: Yes, this is the length in tabular

PR #2, opened 2022-12-16
Update setup.py to adapt repo change

yinweisu:
* Updated setup.py with necessary util functions, and modified them to adapt repo change. Installation has been tested locally.
* Updated dependency version ranges

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,2 +1,2 @@
 Train and deploy AutoGluon backed models on the cloud.
-Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,2 @@
+Train and deploy AutoGluon backed models on the cloud.
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
--- a/setup.py
+++ b/setup.py
@@ -1,50 +1,134 @@
-#!/usr/bin/env python
-###########################
-# This code block is a HACK (!), but is necessary to avoid code duplication. Do NOT alter these lines.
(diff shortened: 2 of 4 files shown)


NOTICE
```diff
@@ -0,0 +1,2 @@
+Train and deploy AutoGluon backed models on the cloud.
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gradientsky, line 2: 2022

src/autogluon/cloud/version.py
```diff
@@ -1,2 +0,0 @@
-"""This is the autogluon version file."""
-__version__ = '0.1b20221111'
```
gradientsky, line 2: This file should be in `.gitignore` because it is auto-generated.

setup.py
```diff
@@ -1,50 +1,134 @@
...
 install_requires = [
-    # version ranges added in ag.get_dependency_version_ranges()
-    "autogluon.common<0.7",
-    "boto3",
-    "numpy",
-    "packaging<22",
-    "pandas",
+    "autogluon.common>=0.6",
+    "boto3<2.0",
+    "numpy>=1.21,<2.0",
```
gradientsky, line 109: Security: `numpy>=1.21.4`

gradientsky:
`CONTRIBUTING` is empty - should be updated since it have important sections about security reporting.

src/autogluon/cloud/version.py
yinweisu: Yea, this is deleting it. I've added it to `.gitignore`

yinweisu:
@gradientsky Will update `CONTRIBUTING.md` in follow-up PRs. Regarding security reporting, should I just reuse what's in `autogluon` `CONTRIBUTING.md`?

gradientsky (approved):

PR #10, opened 2022-12-21
Optimize Tests

yinweisu:
Issue #, if available:

Description of changes:
* General functionalities such as reattach an endpoint doesn't need to be tested for every modality. Hence testing it once, and have each modality to do a basic functionality test to speed up the tests.
* Also removed multimodality test involving only tabular+text as we also test for tabular+text+image, which is a superset of it.
* Added CI badge in readme


gradientsky (approved):

tonyhoo (approved):

suzhoum (commented):
LGTM!

PR #11, opened 2022-12-22
Check multimodal and add support for testing different framework version

yinweisu:
Issue #, if available:

Description of changes:
* Install autogluon tabular/multimodal accordingly to verify trained artifacts did train multimodal models
* Add support for testing of different framework version for the sake of future testing.
* A follow-up PR will be sent involves workflow change.
* Write image to `/tmp` when inference


src/autogluon/cloud/scripts/image_serve.py
```diff
@@ -12,12 +12,15 @@
...
-    for file in files:
-        if file.endswith(".png"):
-            os.remove(file)
+
+def _save_images(im, im_name):
+    os.makedirs(image_dir, exist_ok=True)
+    im_path = os.path.join(image_dir, im_name)
+    im.save(im_path)
+
+    return im_path
```
suzhoum, line 23: Can we make this as an shared util function so that `tabular_serve.py` can also use?

src/autogluon/cloud/scripts/tabular_serve.py
```diff
@@ -10,17 +11,21 @@
...
     im_bytes = base64.b85decode(bytes)
     # nosec B303 - not a cryptographic use
     im_hash = hashlib.sha1(im_bytes).hexdigest()
     im = Image.open(BytesIO(im_bytes))
     im_name = f"tabular_image_{im_hash}.png"
-    im.save(im_name)
-    print(f"Image saved as {im_name}")
+    im_path = os.path.join(image_dir, im_name)
+    im.save(im_path)
+    print(f"Image saved as {im_path}")
```
suzhoum, line 26: Ditto

suzhoum (approved):
LGTM with a minor comment

src/autogluon/cloud/scripts/image_serve.py
yinweisu: No. This are python files that will be sent to the container. Nothing can be shared

suzhoum: Got it. I like that you are making this as a function, maybe we can do the same in `tabular_serve.py`.

PR #13, opened 2022-12-27
Support output content type as list

yinweisu:
Issue #, if available:

Description of changes:
* Support output content type as list
* Disable debug mode
* Fix test using cpu image only

### DIFF
--- a/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -470,7 +470,10 @@ class CloudPredictor(ABC):
             autogluon_sagemaker_estimator_kwargs = {}
         autogluon_sagemaker_estimator_kwargs = copy.deepcopy(autogluon_sagemaker_estimator_kwargs)
         autogluon_sagemaker_estimator_kwargs.pop("output_path", None)
-        if autogluon_sagemaker_estimator_kwargs.get("disable_profiler", None) is None and autogluon_sagemaker_estimator_kwargs.get("debugger_hook_config", None) is None:
+        if (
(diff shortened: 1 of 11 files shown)


tests/conftest.py
```diff
@@ -15,14 +15,22 @@ class CloudTestHelper:
...
-        training_custom_image_uri = CloudTestHelper.cpu_training_image
-        inference_custom_image_uri = CloudTestHelper.cpu_inference_image
+    def get_custom_image_uri(framework_version="source", type="training", gpu=False):
+        assert type in ["training", "inference"]
+        if type == "training":
+            if gpu:
+                custom_image_uri = CloudTestHelper.gpu_training_image
+            else:
+                custom_image_uri = CloudTestHelper.cpu_training_image
+        elif type == "inference":
```
gradientsky, line 25: nit: can be just `else`

src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -470,6 +470,8 @@ def fit(
             autogluon_sagemaker_estimator_kwargs = {}
         autogluon_sagemaker_estimator_kwargs = copy.deepcopy(autogluon_sagemaker_estimator_kwargs)
         autogluon_sagemaker_estimator_kwargs.pop("output_path", None)
+        if autogluon_sagemaker_estimator_kwargs.get("debugger_hook_config", None) is None:
+            autogluon_sagemaker_estimator_kwargs["debugger_hook_config"] = False
```
gradientsky, line 474: Both of these should be added:
```
    disable_profiler=True,
    debugger_hook_config=False,
```

gradientsky (approved):

PR #16, opened 2023-01-05
[Doc] Enable doc CI

yinweisu:
Issue #, if available:

Description of changes:
* The workflow consists two paths:
   1. When a PR submitted, doc will be built and be shared through github bot for review. Example: https://github.com/yinweisu/autogluon-cloud/pull/2
   2. When a push to master/stable branch, the single html page containing the doc will be copied to the main AG website bucket (dev/stable). The main AG website has prepared the TOC through https://github.com/autogluon/autogluon/pull/2634
* Multiple static artifacts are required to maintain the style of the page.
* A follow up PR will be created to add tutorial content

**This won't take effect until merged in**


.github/actions/setup-env-vars/action.yml
```diff
@@ -0,0 +1,34 @@
+name: "Setup Env Vars"
```
tonyhoo, line 1: Is it a duplicate from the AutoGluon? If so, can we reuse the same so we could avoid duplicate codes and changes in 2 places going forward? Something like [this](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow)

docs/config.ini
```diff
@@ -0,0 +1,77 @@
+[project]
+
+# The project name, used as the filename of the package and the PDF file. For
+# example, if set to d2l-book, then will build d2l-book.zip and d2l-book.pdf
+name = autogluon
```
tonyhoo, line 5: shall we use `autogluon.cloud` here?

docs/config.ini
```diff
@@ -0,0 +1,77 @@
...
+
+# Book title. It will be displayed on the top-right of the HTML page and the
+# front page of the PDF file
+title = AutoGluon Documentation
+
+author = AutoGluon contributors
+
+copyright = 2022, All developers. Licensed under the Apache License, Version 2.0.
+
+release = 0.6.2
```
tonyhoo, line 15: Do we follow a different versioning mechanism from cloud module or we want to be always consistent with AG?

docs/requirements_doc.txt
```diff
@@ -0,0 +1,18 @@
+ipykernel>=4.5.1,<6.0.0
+jupyter-sphinx>=0.3.2
+portalocker
+nose
+docutils<0.16
+mu-notedown
+awscli>=1.18.140
```
tonyhoo, line 7: why not awscli 2, new features will only be added to cli 2 now

.github/actions/setup-env-vars/action.yml
yinweisu: Yes, it's duplicated. Not sure if we can use it directly, but I can give it a shot.

docs/config.ini
yinweisu: No, the website we built follow the same metadata as the main autogluon website.
The only thing we need is the cloud page, but it requires the same metadata to avoid conflict when copied over

docs/requirements_doc.txt
yinweisu: We probably can, I just copied this from our main repo

.github/actions/setup-env-vars/action.yml
yinweisu: OK. I just researched it a little bit. What you were referring is `reusable workflow`, while this is a `composite action`. `composite action` cannot be used across different repos.

We might consider replacing this to be a reusable workflow later (if possible), but I'd like to not do it in this PR

tonyhoo: did some research, I think the best way is to create starter [workflow](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization) for common steps for reuse. Fine to proceed with current approach

tonyhoo (approved):

PR #22, opened 2023-01-13
Batch inference `predict_proba`

yinweisu:
Issue #, if available:

Description of changes:
* Add `predict_proba` for batch inference
* Update tutorial
* Fixed a bug in `split_pred_and_pred_proba`


docs/tutorials/autogluon-cloud.md
```diff
@@ -171,9 +171,49 @@ result = cloud_predictor.predict(
...
+```{.python}
+result = cloud_predictor.predict_proba(
+    'test.csv',  # can be a DataFrame as well and the results will be stored in s3 bucket
+    include_predict=True  # Will return a tuple (prediction, prediction probability). Set this to False to get prediction probability only.
+    instance_type="ml.m5.2xlarge",  # Checkout supported instance and pricing here: https://aws.amazon.com/sagemaker/pricing/
+    wait=True,  # Set this to False to make it an unblocking call and immediately return
+    # If True, returns a Pandas Series object of predictions.
+    # If False, returns nothing. You will have to download results separately via cloud_predictor.download_predict_results
+    download=True,
+    persist=True,  # If True and download=True, the results file will also be saved to local disk.
```
Innixma, line 191: What format will the results file be? Will the user know how to parse it? Should we instead save it as 2 files: prediction and pred_proba?

Innixma (approved):
LGTM!

docs/tutorials/autogluon-cloud.md
yinweisu: SageMaker only write a single file to the s3 bucket, and there's no way to differentiate `pred` and `pred_proba` in the inference container. Hence, we write both `pred` and `pred_proba` to s3.

We could do some post-processing after the batch inference to download the file, parse it, and overwrite the file to the s3 bucket instead. But this seems pretty nasty

PR #24, opened 2023-01-17
Code location

yinweisu:
Issue #, if available:

Description of changes:
* Noticed cloud predictor was uploading training script to a different s3 path than the rest of the utils. This PR fix it.
* Have ran a local test and confirmed code has been uploaded to the correct s3 path.

### DIFF
--- a/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -477,6 +477,7 @@ class CloudPredictor(ABC):
             autogluon_sagemaker_estimator_kwargs["disable_profiler"] = True
             autogluon_sagemaker_estimator_kwargs["debugger_hook_config"] = False
         output_path = self.cloud_output_path + "/model"
+        code_location = self.cloud_output_path + "/code"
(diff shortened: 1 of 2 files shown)


gidler (approved):

src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -477,6 +477,7 @@ def fit(
             autogluon_sagemaker_estimator_kwargs["disable_profiler"] = True
             autogluon_sagemaker_estimator_kwargs["debugger_hook_config"] = False
         output_path = self.cloud_output_path + "/model"
+        code_location = self.cloud_output_path + "/code"
```
tonyhoo, line 480: Shall we use `os.path.join` instead?

yinweisu: No. This is S3 path. `os.path.join` would result in `path\code` in Windows and would not work with S3

gradientsky (approved):

src/autogluon/cloud/predictor/cloud_predictor.py
tonyhoo: Thanks for clarification

tonyhoo (approved):

PR #28, opened 2023-01-24
pypi release workflow

yinweisu:
Issue #, if available:

Description of changes:
* Modified based on the workflow being used by main AG repo

### DIFF
--- a/.github/workflows/pypi_release.yml
+++ b/.github/workflows/pypi_release.yml
@@ -15,9 +15,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v3.3.0
     - name: Set up Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v4.5.0
       with:
         python-version: '3.8'
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python
+      uses: actions/setup-python@v1
```
tonyhoo, line 20: I thought it is not the latest stable version anymore. Can you check if we need to upgrade here (e.g. for security reasons)?

yinweisu: You correct. These are not the latest. Let me update them to be latest and we probably want to do the same to AG workflow as well

tonyhoo: Thanks

tonyhoo (approved):

.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,34 @@
...
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3.3.0
```
gradientsky, line 18: Can we just do `@v3` as it is highlighted in tutorials on the [action github](https://github.com/actions/checkout)?

.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4.5.0
```
gradientsky, line 20: ditto: `@v4`

.github/workflows/pypi_release.yml
yinweisu: Well, if you click use latest button on this page: https://github.com/marketplace/actions/checkout you get what I put there. I think it doesn't matter to much anyways...

.github/workflows/pypi_release.yml
yinweisu: ditto

gradientsky (approved):

PR #38, opened 2023-02-03
Prepare timeseries cloud ci

yinweisu:
Issue #, if available:

Description of changes:


.github/workflows/continuous_integration.yml
```diff
@@ -217,9 +217,40 @@ jobs:
         with:
           submodule-to-test: multimodal
           ag_version: '${{ matrix.AG_VERSION }}'
+  test_timeseries_cloud:
+    strategy:
+      matrix:
+        AG_VERSION: ["source", "0.6.1"]
```
tonyhoo, line 223: We shall add check for `0.6.2` as well. Meanwhile, shall we define a global variable on the env level so we don't need to modify it each time a new version published

tonyhoo (approved):

PR #47, opened 2023-03-01
Ray AWS Cluster Config Generator

yinweisu:
Issue #, if available:

Description of changes:
* Implemented `RayAWSClusterConfigGenerator`. It has a default config and accepts user config override. Also allow users to update the config with frequent used parameters that are available in `fit` call, i.e. `instance_type`


src/autogluon/cloud/cluster/cluster_config_generator.py
```diff
@@ -1,19 +1,65 @@
...
+import yaml
+
+DEFAULT_CONFIG_LOCATION = os.path.join(os.path.dirname(__file__), "..", "default_cluster_configs")
 
 
 class ClusterConfigGenerator(ABC):
-    @abstractmethod
-    @staticmethod
-    def get_default_config() -> Dict[str, Any]:
+    default_config_file = os.path.join(DEFAULT_CONFIG_LOCATION, "DUMMY")
```
gradientsky, line 11: Not sure we should use `DUMMY` value for the default location; can be `DEFAULT_CONFIG` or something else neutral and descriptive.

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
```diff
@@ -0,0 +1,146 @@
...
+        region, Optional[str]
+            Region to launch the cluster. Default us-east-1
+        """
+        super().__init__(config=config)
+        self._update_region(region)
+
+    def _update_config(
+        self,
+        instance_type: Optional[str] = None,
+        instance_count: Optional[str] = None,
```
gradientsky, line 44: Why not set it to 1 as a default and change type to `int` in the API?
`instance_count: int = 1` ?

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
```diff
@@ -0,0 +1,146 @@
...
+    DOCKER,
+    EBS,
+    IMAGE,
+    INSTANCE_TYPE,
+    MAX_WORKERS,
+    MIN_WORKERS,
+    NODE_CONFIG,
+    PROVIDER,
+    REGION,
+    VOLUMN_SIZE,
```
gradientsky, line 17: `VOLUME_SIZE`?

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
```diff
@@ -0,0 +1,146 @@
...
+                https://docs.ray.io/en/latest/cluster/vms/references/ray-cluster-configuration.html#available-node-types
+        custom_image_uri: str, default = None
+            Custom image to be used by the cluster container. The image MUST have Ray and AG installed.
+            If provided, will overwrite `docker.image`
+            To learn more,
+                https://docs.ray.io/en/latest/cluster/vms/references/ray-cluster-configuration.html#docker-image
+        """
+        self._update_instance_type(instance_type=instance_type)
+        self._update_instance_count(instance_count=instance_count, worker_node_name=worker_node_name)
+        self._update_volumn_size(volumes_size=volumes_size)
```
gradientsky, line 84: `_update_volume_size`?

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
```diff
@@ -0,0 +1,146 @@
...
+            available_node_types = default_config[AVAILABLE_NODE_TYPES]
+        self.config.update(available_node_types)
+
+    def _set_provider(self):
+        """Set provider to be default ones if user didn't provide any"""
+        default_config = self._default_config
+        provider: Dict[str, Any] = self.config.get(PROVIDER, None)
+        if provider is None:
+            provider = default_config[PROVIDER]
+        self.config.update({PROVIDER: provider})
```
gradientsky, line 101: Can we templatize repeating methods: I see lots of repeating code which can be reuse between the `_update_xxx` methods.

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
```diff
@@ -0,0 +1,146 @@
...
+            worker_instance_count = instance_count - 1
+            assert worker_instance_count >= 0
+            self.config[MAX_WORKERS] = worker_instance_count
+            self._set_available_node_types()
+            assert (
+                worker_node_name in self.config[AVAILABLE_NODE_TYPES]
+            ), f"Didn't find node definition for {worker_node_name}. Please make sure you provided the correct `worker_node_name`"
+            self.config[AVAILABLE_NODE_TYPES][worker_node_name].update({MIN_WORKERS: worker_instance_count})
+
+    def _update_volumn_size(self, volumes_size):
```
gradientsky, line 128: volume

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
yinweisu: It should be int, that's a mistake. Though if we specify a default value for instance_count, we will always update the config to have instance_count = 1 even when user left it blank. I'll stick with default = None, meaning user doesn't want to update it

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
yinweisu: It is kind of hard to templatize these methods because they are essentially trying to update keys and values lying in different layers of the yaml file. The method looks similar, but because the structure of the yaml file differs for each key, it's actually specific

PR #48, opened 2023-03-06
Add cluster role to CI

yinweisu:
Issue #, if available:

Description of changes:

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -271,11 +271,11 @@ jobs:
             other_than_docs:
               - '!(docs/**)**'
       - name: Configure AWS Credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -270,6 +270,12 @@ jobs:
           filters: |
             other_than_docs:
               - '!(docs/**)**'
+      - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
```
gidler, line 275: Is the indention correct on these lines?

yinweisu: Nice catch! Fixed

gidler (approved):

PR #49, opened 2023-03-07
RayAWSClusterManager

yinweisu:
Issue #, if available:

Description of changes:
* Implement `RayAWSClusterManager`. Mostly a wrapper around ray command. The dashboard command is a blocking call that will block the whole terminal; hence used nohup command as a workaround
* Add default iam policy and trust relationship template
  * Moved sagemaker iam policy and trust relationship to an individual file for better structure
* Added `save_config` to `ClusterConfigGenerator`
* Added `cluster_name` in constructor of `RayAWSClusterConfigGenerator`
* Added method to allow user to provide instance profile to be used by head and worker nodes


tests/unittests/cluster/test_ray_cluster.py
```diff
@@ -0,0 +1,23 @@
+import os
+import tempfile
+import time
+
+from autogluon.cloud.cluster import RayAWSClusterConfigGenerator, RayAWSClusterManager
+
+
+def test_ray_aws_cluster(test_helper):
```
tonyhoo, line 8: wondering what we are testing here. Shall we verify no error thrown for `up()` and `setup_connection()` or we should really test the connection is working

src/autogluon/cloud/default_cluster_configs/ray_aws_default_cluster_config.yaml
```diff
@@ -2,12 +2,13 @@
 cluster_name: ag_ray_aws_default
 
 # The maximum number of workers nodes to launch in addition to the head
-max_workers: 2
+max_workers: 1
```
tonyhoo, line 5: why we reduce it to 1?

tests/unittests/cluster/test_ray_cluster.py
yinweisu: Yea. We mainly test the `up()` and `setup_connection()` doesn't throw error now. Once the `RayJob` interface is added, we'll test the connection is actually working by submitting a job.

src/autogluon/cloud/default_cluster_configs/ray_aws_default_cluster_config.yaml
yinweisu: This is just the default config, I set it to 1 to minimize cost. Though the default config will never be used actually because the cloud predictor will set the worker value to a custom one eventually

PR #51, opened 2023-03-08
RayJob

yinweisu:
Issue #, if available:

Description of changes:
* Implemented RayJob interface
* Tests related to job submission involves http request to local host, which is blocked by github action. Marked those tests as local only, and tested locally


src/autogluon/cloud/job/ray_job.py
```diff
@@ -27,20 +61,79 @@ def info(self) -> Dict:
...
+        if ray_submit_job_args is None:
+            ray_submit_job_args = {}
+        self.client.submit_job(
+            entrypoint=entry_point, runtime_env=runtime_env, submission_id=job_name, **ray_submit_job_args
+        )
+        logger.log(20, f"Submitted job {job_name} to the cluster")
+        self._job_name = job_name
+        if wait:
+            logger.warning("We don't support log streaming currently. Logs will be avaialbe when the job is finished")
+            self._wait_until_status(
```
tonyhoo, line 108: Not sure if there is an timeout mechanism on ray job but would be good to add a timeout instead of waiting forever

tonyhoo (approved):

PR #62, opened 2023-03-27
Update Web Style

yinweisu:
Issue #, if available:

Description of changes:
* Update web style to match the new style in main AG web
* Disabled cluster test for now as it triggers some weird error on github action. This feature will not be available in 0.1.1 hence ok to disable to unblock release


docs/conf.py
```diff
@@ -0,0 +1,83 @@
...
+
+# See https://myst-parser.readthedocs.io/en/latest/syntax/optional.html
+myst_enable_extensions = ['colon_fence', 'deflist', 'dollarmath', 'html_image', 'substitution']
+
+autosummary_generate = True
+numpydoc_show_class_members = False
+
+googleanalytics_id = "UA-96378503-20"
+
+nb_execution_mode = 'force'
```
gidler, line 32: You can set this to `off` while you don't have any notebooks that need to be executed.

docs/conf.py
```diff
@@ -0,0 +1,83 @@
...
+
+nb_dirs_to_exec = [os.path.join('tutorials', tag) for tag in tags if os.path.isdir(os.path.join('tutorials', tag))]
+
+if len(nb_dirs_to_exec) > 0:
+    nb_dirs_to_exclude = [dirpath for dirpath, _, filenames in os.walk('tutorials')
+                            if any(map(lambda x: x.endswith('.ipynb'), filenames))
+                               and not dirpath.startswith(tuple(nb_dirs_to_exec))]
+
+    for nb_dir in nb_dirs_to_exclude:
+        nb_execution_excludepatterns.append(os.path.join(nb_dir, '*.ipynb'))
```
gidler, line 47: If you don't need the ability to exclude building some notebooks, I like the idea of getting rid of these lines because they take extra time to understand.

docs/index.md
```diff
@@ -0,0 +1,91 @@
...
+test_data = pd.read_csv("https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv")
+predictor_init_args = {"label": "class"}  # init args you would pass to AG TabularPredictor
+predictor_fit_args = {"train_data": train_data, "time_limit": 120}  # fit args you would pass to AG TabularPredictor
+cloud_predictor = TabularCloudPredictor(cloud_output_path='YOUR_S3_BUCKET_PATH')
+cloud_predictor.fit(predictor_init_args=predictor_init_args, predictor_fit_args=predictor_fit_args)
+cloud_predictor.deploy()
+result = cloud_predictor.predict_real_time(test_data)
+cloud_predictor.cleanup_deployment()
+# Batch inference
+result = cloud_predictor.predict(test_data)
```
gidler, line 56: Having no new-lines here makes this a bit hard to read through. Could we run it through `black`?

src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -650,6 +650,7 @@ def predict_proba(
         return self.backend.predict_proba(
             test_data=test_data,
             test_data_image_column=test_data_image_column,
+            include_predict=include_predict,
```
gidler, line 653: Just curious, where did this come from?

yinweisu: I just noticed this variable is not being passed in after the backend update so decided to add it here

docs/conf.py
yinweisu: I think we can leave it as is cause it's possible that we will want to execute some content in the future, and it would be confusing for someone unfamiliar with this to know how to enable execution

docs/conf.py
yinweisu: I think this is a nice feature and could be useful in the future so I left it here

gidler (approved):

PR #72, opened 2023-04-18
Tabular Distributed Training

yinweisu:
Issue #, if available:

Description of changes:
* Add tabular distributed training main functionality
* Following PRs will be added to refine details, i.e. auto shut-down the cluster when job finishes if `wait = False`
* Updated cluster unit tests
* Benchmark was conducted to verify functionality and stability

Example usage:
```python
from autogluon.cloud import TabularCloudPredictor

import pandas as pd
import time


cp = TabularCloudPredictor(
     ...
     backend = "ray_aws"
)

train_data = "train_data"
predictor_init_args = {
    "label": "label"
}  # init args you would pass to AG TabularPredictor
predictor_fit_args = {
    "train_data": train_data,
    "num_bag_folds": 8,
    "num_bag_sets": 1
}  # fit args you would pass to AG TabularPredictor

cp.fit(
    predictor_init_args=predictor_init_args,
    predictor_fit_args=predictor_fit_args
)
```


src/autogluon/cloud/utils/ray_aws_iam.py
```diff
@@ -47,20 +51,48 @@
...
                 "iam:CreateRole",
+                "iam:CreatePolicy",
+                "iam:DeletePolicy",
                 "iam:GetRole",
                 "iam:AttachRolePolicy",
+                "iam:DetachRolePolicy",
                 "iam:AddRoleToInstanceProfile",
                 "iam:PassRole",
             ],
             "Resource": ["*"],
```
gradientsky, line 70: The `CreatePolicy`/`DeletePolicy` roles are too powerful to give permissions like these: is it possible to narrow-down `Resource` to something more specific than `*`?

src/autogluon/cloud/cluster/ray_cluster_manager.py
```diff
@@ -89,7 +90,7 @@ def setup_dashboard(self, port: int = 8265, block: bool = False, **kwargs) -> No
         """
         cmd = f"ray dashboard -p {port} {self.config}"
         if not block:
-            cmd = "nohup " + cmd + "&>/dev/null &"
+            cmd = "nohup " + cmd + " >/dev/null 2>&1 &"
```
gradientsky, line 93: not safe use of CLI: allows command injection: i.e. pass port to something like `123 | rm -rf some_dir`

src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -29,7 +29,11 @@ class CloudPredictor(ABC):
     predictor_file_name = "CloudPredictor.pkl"
 
     def __init__(
-        self, cloud_output_path: str, local_output_path: Optional[str] = None, backend=SAGEMAKER, verbosity: int = 2
+        self,
+        cloud_output_path: str,
+        local_output_path: Optional[str] = None,
+        backend: str = SAGEMAKER,
```
gradientsky, line 35: Please add validation if the value is in specified set of values.

src/autogluon/cloud/utils/ray_aws_iam.py
yinweisu: Good point. I'll restrict these two to follow a specific pattern

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
```diff
@@ -101,21 +115,29 @@ def _update_config(
             The instance profile arn to be attched to the worker node. If not specified, a default one will be created with full EC2 and S3 permissions.
             To learn how to scope down the permission,
                 https://github.com/ray-project/ray/issues/9327
+        initialization_commands: Optional[List[str]], default = None
+            The initialization commands of the ray cluster.
+            If not specified, will contain a default ECR login command to be able to pull AG DLC image, i.e.
+                - aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin 763104351884.dkr.ecr.us-east-1.amazonaws.com
```
tonyhoo, line 121: Do we know which version of aws cli is installed. The command slightly differs between v1 and v2

src/autogluon/cloud/default_cluster_configs/ray_aws_default_cluster_config.yaml
```diff
@@ -4,6 +4,11 @@ cluster_name: ag_ray_aws_default
 # The maximum number of workers nodes to launch in addition to the head
 max_workers: 1
 
+docker:
+    # The image uri will be dynamically replaced by cloud predictor
+    image: "763104351884.dkr.ecr.us-east-1.amazonaws.com/autogluon-training:0.7.0-cpu-py39-ubuntu20.04"
```
tonyhoo, line 9: Shall we have a different default for GPU instance?

src/autogluon/cloud/scripts/ray_scripts/train.py
```diff
@@ -0,0 +1,106 @@
...
+        prefix = None
+    if prefix is not None:
+        object_name = prefix + "/" + object_name
+
+    # Upload the file
+    s3_client = boto3.client("s3")
+    s3_client.upload_file(file_name, bucket, object_name)
+
+
+if __name__ == "__main__":
```
tonyhoo, line 63: [minor suggestion] we can use [typer](https://typer.tiangolo.com/) to simplify this significantly

src/autogluon/cloud/cluster/ray_cluster_manager.py
yinweisu: This command will be called by the user themselves and is not exposed inside the cluster. (There's no autogluon.cloud being installed inside the cluster). Therefore, I think this should be fine

.github/workflow_scripts/env_setup.sh
```diff
@@ -36,6 +36,9 @@ function install_latest_tabular {
 function install_latest_multimodal {
     install_latest_tabular_and_multimodal_dependencies
     python3 -m pip install -e autogluon/multimodal/
+    mim install mmcv-full --timeout 60
+    python3 -m pip install --upgrade "mmdet>=2.28, <3.0.0"
```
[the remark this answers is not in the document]
yinweisu: Had a look at it, I think it's just a template that allow you to build a new workflow faster. It's not something that will be ran everytime at the start

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
yinweisu: This is actually a good point. I think it is whatever being installed on ec2 by default. This command is being called before we enter the docker container.
When I look at the instruction in ECR repo, this is the only command that it shows for login. I assume it's ok for now.

src/autogluon/cloud/default_cluster_configs/ray_aws_default_cluster_config.yaml
yinweisu: This is just a place holder. We'll replace it with GPU one if needed.

src/autogluon/cloud/scripts/ray_scripts/train.py
yinweisu: This is a good point. Though the issue here is that this script will be ran in the DLC container, meaning `typer` has to be a dependency of AG. I think it's probably not necessary to introduce such a dependency when AG itself doesn't need it

src/autogluon/cloud/cluster/ray_aws_cluster_config_generator.py
tonyhoo: Please make sure the AMI used by the EC2 is security blessed

tonyhoo: Sounds good to me. Thanks for following up

tonyhoo (approved):

PR #84, opened 2023-08-11
Add volume deploy option

yinweisu:
Issue #, if available:

Description of changes:
* Allow user to set volume size when deploy. Also add check for gpu instances as volume size is not supported


suzhoum (approved):
LGTM!

PR #110, opened 2024-05-08
0.4.0

suzhoum:
Issue #, if available:

Description of changes:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,29 +104,29 @@ def default_setup_args(*, version):
     return setup_args
 
 
-version = "0.3.1"
+version = "0.4.0"
 version = update_version(version, use_file_if_exists=False, create_file=True)
 
 install_requires = [
     # common module provides utils with stable api across minor version
-    "autogluon.common>=0.7,<1.1",
+    "autogluon.common>=0.7",
     # <2 because unlikely to introduce breaking changes in minor releases. >=1.10 because 1.10 is 3 years old, no need to support older
     "boto3>=1.10,<2.0",
-    "numpy>=1.21,<1.27",
+    "numpy>=1.21,<1.29",
     "packaging>=23.0,<24.0",
-    "pandas>=2.0.0,<2.2.0",
+    "pandas>=2.0.0,<2.3.0",
     # updated sagemaker is required to fetch latest container info, so we don't want to cap the version too strict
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -104,37 +104,37 @@ def default_setup_args(*, version):
...
 version = update_version(version, use_file_if_exists=False, create_file=True)
 
 install_requires = [
     # common module provides utils with stable api across minor version
-    "autogluon.common>=0.7,<1.1",
+    "autogluon.common>=0.7",
     # <2 because unlikely to introduce breaking changes in minor releases. >=1.10 because 1.10 is 3 years old, no need to support older
     "boto3>=1.10,<2.0",
-    "numpy>=1.21,<1.27",
+    "numpy>=1.21,<1.29",
```
tonyhoo, line 115: We don't need this as core comes with numpy dependency: https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py#L20

setup.py
```diff
@@ -104,37 +104,37 @@ def default_setup_args(*, version):
...
     "packaging>=23.0,<24.0",
-    "pandas>=2.0.0,<2.2.0",
+    "pandas>=2.0.0,<2.3.0",
```
tonyhoo, line 117: Same here

setup.py
```diff
@@ -104,37 +104,37 @@ def default_setup_args(*, version):
...
     # updated sagemaker is required to fetch latest container info, so we don't want to cap the version too strict
     # otherwise cloud module needs to be released to support new container
     "sagemaker>=2.126.0,<3.0",
     "pyarrow>=11.0,<11.1",
     "PyYAML~=6.0",
-    "Pillow>=9.3.0,<10.0",  # unlikely to introduce breaking changes in minor releases
-    "ray[default]>=2.6.3,<2.7",
+    "Pillow>=10.2,<11",  # unlikely to introduce breaking changes in minor releases
```
tonyhoo, line 123: same here

setup.py
suzhoum: Good point. Will update!

setup.py
suzhoum: I kept this as these are not dependencies in `autogluon.common`

tonyhoo (approved):

Issue #130, opened 2024-07-11
Not compatible with latest sagemaker


tonyhoo:
With the release of SageMaker, a new field, `model_reference_arn`, was introduced. Refer to this [PR](https://github.com/aws/sagemaker-python-sdk/pull/4748) for more details. However, it caused our integration [tests](https://github.com/autogluon/autogluon-cloud/actions/runs/9872975903) to fail. The error message displayed was `FAILED tests/unittests/image/test_image.py::test_multimodal_image_only - TypeError: AutoGluonNonRepackInferenceModel.prepare_container_def() got an unexpected keyword argument 'model_reference_arn'`.

PR #131, opened 2024-07-11
add model_reference_arn in signature

suzhoum:
Issue #, if available:
Addressed two issues:
1. https://github.com/autogluon/autogluon-cloud/issues/130
2. Only create policy when not already existed due to restrictions on "Permission Mutation Policies" and we will lose policy that allows us to create new roles.

Description of changes:

### DIFF
--- a/src/autogluon/cloud/utils/ag_sagemaker.py
+++ b/src/autogluon/cloud/utils/ag_sagemaker.py
@@ -190,6 +190,7 @@ class AutoGluonRepackInferenceModel(AutoGluonSagemakerInferenceModel):
         accelerator_type=None,
         serverless_inference_config=None,
         accept_eula=None,
+        model_reference_arn=None,
     ):  # pylint: disable=unused-argument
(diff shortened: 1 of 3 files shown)


tonyhoo (approved):
Thanks for the quick fix!

src/autogluon/cloud/utils/ag_sagemaker.py
prateekdesai04, line 1: Should this change also propagate to the `AutoGluonSagemakerInferenceModel` class followed by wherever deploy is called in `cloud_predictor.py` and `sagemaker_backend.py` ?

suzhoum: This might not be necessary. My take is that we need to align all args under the function `prepare_container_def` with the sagemaker definition here https://sagemaker.readthedocs.io/en/stable/api/inference/model.html#sagemaker.jumpstart.model.JumpStartModel.prepare_container_def, but we are not really using any of them inside the function.

PR #65, opened 2023-03-28
Add 0.1.1 Release Note

yinweisu:
Issue #, if available:

Description of changes:

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #138, opened 2024-09-17
Fix batch transform issue for tabular predictor with multiple partitions

tonyhoo:
### Description:
This PR fixes the issue where batch transform jobs fail due to column misalignment when the input CSV file is partitioned into multiple records. The problem arises because headers from different partitions are not handled properly, leading to misaligned columns and prediction failures during inference.

**Changes:**
- Added logic to align columns across partitions by ensuring headers are managed correctly.
- Introduced `_read_with_fallback` and `_align_columns` helper functions to handle column alignment.
- Updated `transform_fn` in `tabular_serve.py` to use these helper functions.

**Limitations:**
- This fix currently only works for the tabular predictor. Support for multimodal and timeseries predictors depends on the implementation of `original_features`, which can be tracked in [issue #4477](https://github.com/autogluon/autogluon/issues/4477).

**Steps to Reproduce:**
The following script can be used to reproduce the issue:

```python
from autogluon.cloud import TabularCloudPredictor
import pandas as pd

# Load datasets
train_data = pd.read_csv("https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv")
test_data = pd.read_csv("https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv")
test_data.drop(columns=['class'], inplace=True)

# Cloud Predictor Arguments
predictor_init_args = {"label": "class"}  
predictor_fit_args = {"train_data": train_data, "time_limit": 60}  

# Initialize Cloud Predictor and Fit
cloud_predictor = TabularCloudPredictor(cloud_output_path='tonyhu-autogluon')
cloud_predictor.fit(predictor_init_args=predictor_init_args, predictor_fit_args=predictor_fit_args)

# Batch Inference with small max_payload to force multiple partitions
result = cloud_predictor.predict(test_data, backend_kwargs={"transformer_kwargs": {"max_payload": 1}})
```

**Expected Behavior:**
The batch transform job should handle multiple partitions correctly, aligning columns across the partitions and ignoring or managing headers if present in individual partitions.

**Observed Behavior:**
The job fails with the following error logs:
```
Bad HTTP status received from algorithm: 500
invalid literal for int() with base 10: '0.1': Error while type casting for column 'capital-loss'
```

Logs show that the columns are misaligned for certain partitions:
```
test_columns: [' 11th', ' Machine-op-inspct', ' Male', ' Never-married', ' Other-relative', ' Private', ' United-States', ' White', '0', '0.1', '207443', '50', '62', '7']
2024-09-13T21:56:19,062 [INFO ] W-9000-model_1.0-stdout MODEL_LOG - train_columns: ['age', 'capital-gain', 'capital-loss', 'education', 'education-num', 'fnlwgt', 'hours-per-week', 'marital-status', 'native-country', 'occupation', 'race', 'relationship', 'sex', 'workclass']
```

**Environment:**
- `autogluon==1.1.0`
- Running batch transform in SageMaker with `MultiRecord` strategy.
- `MaxPayloadInMB=1` is set to ensure multiple partitions.

**Additional Information:**
The issue seems to be that AutoGluon Cloud is not handling the headers properly when dealing with batch transform partitioned records. In a multi-partition job, not all batches will have the header/column, which is causing the column misalignment.

**Note:**
This fix currently only works for the tabular predictor. Support for multimodal and timeseries predictors depends on the implementation of `original_features`, which can be tracked in [issue #4477](https://github.com/autogluon/autogluon/issues/4477).

### DIFF
--- a/src/autogluon/cloud/backend/sagemaker_backend.py
+++ b/src/autogluon/cloud/backend/sagemaker_backend.py
@@ -319,7 +319,7 @@ class SagemakerBackend(Backend):
             leaderboard=leaderboard,
         )
         # Get the label from predictor_init_args
-        label = predictor_init_args.get('label') or predictor_init_args.get('target') or None
+        label = predictor_init_args.get("label") or predictor_init_args.get("target") or None
         if image_column is not None:
             ag_args["image_column"] = image_column
         ag_args_path = os.path.join(self.local_output_path, "utils", "ag_args.pkl")
@@ -1189,9 +1189,6 @@ class SagemakerBackend(Backend):
             else:
                 test_data = load_pd.load(test_data)
 
-        # Ensure columns are in sync with model expectations for tabular use cases
-        # TODO: Add support for other modalities once features are exposed in APIs
-        # Tracking at https://github.com/autogluon/autogluon/issues/4477
         if isinstance(test_data, pd.DataFrame) and original_features is not None:
             expected_columns = original_features
             incoming_columns = test_data.columns.tolist()
--- a/src/autogluon/cloud/backend/sagemaker_backend.py
+++ b/src/autogluon/cloud/backend/sagemaker_backend.py
@@ -1006,6 +1006,8 @@ class SagemakerBackend(Backend):
                 )
 
         train_input = train_data
+        if isinstance(train_data, str):
+            train_data = load_pd.load(train_data)
         self.original_features = [col for col in train_data.columns if col != label]
         train_data = self._prepare_data(train_data, "train")
         logger.log(20, "Uploading train data...")
--- a/src/autogluon/cloud/backend/sagemaker_backend.py
+++ b/src/autogluon/cloud/backend/sagemaker_backend.py
@@ -902,7 +902,7 @@ class SagemakerBackend(Backend):
             model_kwargs=model_kwargs,
             transformer_kwargs=transformer_kwargs,
             transform_kwargs=transform_kwargs,
-            local_predictor=local_predictor, 
+            local_predictor=local_predictor,
         )
 
         if include_predict:
@@ -1112,7 +1112,6 @@ class SagemakerBackend(Backend):
             raise e
 
     def _upload_batch_predict_data(self, test_data, bucket, key_prefix):
-
         if isinstance(test_data, pd.DataFrame):
             test_data = self._prepare_data(test_data, "test", output_type="csv")
         logger.log(20, "Uploading data...")
--- a/src/autogluon/cloud/backend/sagemaker_backend.py
+++ b/src/autogluon/cloud/backend/sagemaker_backend.py
@@ -710,6 +710,7 @@ class SagemakerBackend(Backend):
         model_kwargs: Optional[Dict] = None,
         transformer_kwargs: Optional[Dict] = None,
         transform_kwargs: Optional[Dict] = None,
+        local_predictor=None,
     ) -> Optional[pd.Series]:
         """
         Predict using SageMaker batch transform.
@@ -766,6 +767,8 @@ class SagemakerBackend(Backend):
             Any extra arguments needed to pass to transform.
             Please refer to
             https://sagemaker.readthedocs.io/en/stable/api/inference/transformer.html#sagemaker.transformer.Transformer.transform for all options.
+        local_predictor: Optional[Any], default = None
+            Local predictor instance to use for batch inference
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -1104,15 +1112,6 @@ def _predict_real_time(self, test_data, accept, split_pred_proba=True, inference
...
     def _upload_batch_predict_data(self, test_data, bucket, key_prefix):
-        # If a directory of images, upload directly
-        if isinstance(test_data, str) and not os.path.isdir(test_data):
-            # either a file to a dataframe, or a file to an image
-            if is_image_file(test_data):
-                logger.warning(
-                    "Are you sure you want to do batch inference on a single image? You might want to try `deploy()` and `predict_real_time()` instead"
-                )
-            else:
-                test_data = load_pd.load(test_data)
```
suzhoum, line 1115: Will it cause problem if images are uploaded here?

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -1179,6 +1179,27 @@ def _predict(
             test_data = convert_image_path_to_encoded_bytes_in_dataframe(
                 dataframe=test_data, image_column=test_data_image_column
             )
+
+        # If a directory of images, upload directly
+        if isinstance(test_data, str) and not os.path.isdir(test_data):
+            # either a file to a dataframe, or a file to an image
+            if is_image_file(test_data):
+                raise ValueError("Image file is not supported for batch inference")
```
suzhoum: Maybe reword to "Single image file"? And suggest to try `.deploy()`

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -1268,8 +1289,8 @@ def _predict(
             pred = results
             if split_pred_proba:
                 pred, pred_proba = split_pred_and_pred_proba(results)
-        if not persist:
-            os.remove(results_path)
+            if not persist:
+                os.remove(results_path)
```
suzhoum, line 1293: Good catch!

src/autogluon/cloud/scripts/sagemaker_scripts/tabular_serve.py
```diff
@@ -40,60 +40,50 @@ def model_fn(model_dir):
...
             break
-    # save image column bytes to disk and update the column with saved path
+
     if image_column is not None:
         print(f"Detected image column {image_column}")
         data[image_column] = [_save_image_and_update_dataframe_column(bytes) for bytes in data[image_column]]
 
+    # Ensure inference_kwargs is a dictionary right before use
+    if inference_kwargs is None:
+        inference_kwargs = {}
```
suzhoum, line 84: The check seems to be redundant?

src/autogluon/cloud/backend/sagemaker_backend.py
tonyhoo: These logics have been copied over to line #L1184 - #L1192

src/autogluon/cloud/scripts/sagemaker_scripts/tabular_serve.py
tonyhoo: No, this code are moved from previous #L64 where to make sure if `inference_kwargs` are explicitly set up to None in payload, we will still have the dict format

src/autogluon/cloud/backend/sagemaker_backend.py
suzhoum: I figured. Now after I read it again, I think moving this to `_predict()` makes sense.

src/autogluon/cloud/scripts/sagemaker_scripts/tabular_serve.py
suzhoum: I think you updated line #65 to make the default value as `{}`, so `inference_kwargs` should not be `None`. But this is a good safety nest.

suzhoum (approved):
Thanks for the fix!

src/autogluon/cloud/scripts/sagemaker_scripts/tabular_serve.py
tonyhoo: The default will be set to `{}` only if `inference_kwargs ` not available from `payload`, it will be set to `None` if 
```
payload = {
    "inference_kwargs": None
}
```

Issue #116, opened 2024-05-31
Allow passing future values of `known_covariates` to `TimeSeriesCloudPredictor.predict` 


shchur:
Currently [predict](https://auto.gluon.ai/cloud/stable/api/autogluon.cloud.TimeSeriesCloudPredictor.predict.html) and [predict_real_time](https://auto.gluon.ai/cloud/stable/api/autogluon.cloud.TimeSeriesCloudPredictor.predict_proba_real_time.html) methods of the TimeSeriesCloudPredictor do not accept the future values of known covariates, which is a major difference from the `TimeSeriesPredictor` API.  We should allow users to pass future values of known covariates at predict time.

3atshan:
Hi any updates on this? predict_real_time does accept kwargs, but they are not actually passed through to the function.

PR #160, opened 2024-11-20
Update timeseries_cloud_predictor.py

3atshan:
Issue #116

Description of changes:
Simply pass through **kwargs from the method call to its internal method call, which will allow passing through inference_kwargs that may be required for inference (for example future_covariates).


suzhoum:
Completed by https://github.com/autogluon/autogluon-cloud/pull/181

Issue #70, opened 2023-03-31
[Feature Request] TimeSeriesCloudPredictor known_covariates


austinmw:
It seems that currently `TimeSeriesCloudPredictor` training already supports known covariates by adding these to the training data frame and additionally including `predictor_init_args["known_covariates_names"]`.

However `predict_real_time` and `predict` methods won't accept the additional/necessary `known_covariates` argument that needs to be passed for inference as [described here](https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-indepth.html#time-varying-covariates).

Please add as this is an important forecasting feature 😁

PR #113, opened 2024-05-22
Support kwargs in predict() and predict_proba()

suzhoum:
Issue #, if available:

Description of changes:

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #181, opened 2025-04-28
Enable known_covariates in TimeSeriesCloudPredictor.predict_real_time

### LINKED ISSUE #116: Allow passing future values of `known_covariates` to `TimeSeriesCloudPredictor.predict`

shchur:
Issue #, if available: Fixes #116, #70

Description of changes:
- Forward the `known_covariates` kwarg to `TimeSeriesCloudPredictor.predict_real_time` via the `inference_kwargs` to the `TimeSeriesSagemakerBackend.predict_real_time` method.
- Add a comment explaining that currently `known_covariates` cannot be used during batch prediction (only real-time inference supported)
- Updated stale pointers to the AutoGluon documentation.

### DIFF
--- a/src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
+++ b/src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
@@ -80,7 +80,7 @@ class TimeSeriesCloudPredictor(CloudPredictor):
         static_features: Optional[pd.DataFrame]
              An optional data frame describing the metadata attributes of individual items in the item index.
              For more detail, please refer to `TimeSeriesDataFrame` documentation:
-             https://auto.gluon.ai/stable/api/autogluon.predictor.html#timeseriesdataframe
+             https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesDataFrame.html
         framework_version: str, default = `latest`
             Training container version of autogluon.
             If `latest`, will use the latest available container version.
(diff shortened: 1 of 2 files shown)


src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -198,6 +204,7 @@ def predict_real_time(
             target=self.target_column,
             static_features=static_features,
             accept=accept,
+            inference_kwargs=dict(known_covariates=known_covariates, **kwargs),
```
shchur, line 207: I validated that this works locally but couldn't update the [unit tests](https://github.com/autogluon/autogluon-cloud/blob/master/tests/unittests/timeseries/test_timeseries.py#L47-L52) because currently `known_covariates` can only work with real-time prediction (and not batch prediction). One possible solution would be to train two predictors: one that expects `known_covariates` and one that doesn't, but I think that would be too much of an overhead.

suzhoum: We might want to revisit this PR https://github.com/autogluon/autogluon-cloud/pull/113 for batch prediction support.

suzhoum (approved):
LGTM!

PR #210, opened 2026-05-08
Disable AutoMM tests

shchur:
Issue #, if available:

Description of changes:
- Only test with AutoGluon v1.5.0 images - no source install
- skip AutoMM and Ray tests


AnirudhDagar (approved):
thanks, going forward do we want to not support any MultiModal users (given AG MultiModal is not being maintained)?

PR #213, opened 2026-05-08
Add autogluon-cloud python setup API

AnirudhDagar:
Introduces `autogluon.cloud.bootstrap/register/status/teardown` Python API. This functionality will be extended to a CLI interface as well using `click` and `rich` in a subsequent PR. Both will use the shared `cloud_setup` engine that provisions the CloudFormation stack, writes a per-profile config at ~/.autogluon/cloud.yaml, and tears it down cleanly.

## Usage

  ```python
  from autogluon.cloud import bootstrap, status, teardown

  bootstrap()       # uses the defaults
  # OR
  bootstrap(backend="sagemaker", session=<boto3 session>, stack_name="my_stack")
  
  status()                                    # health check
  teardown(delete_bucket_contents=True)       # cleanup
  ```

  `bootstrap()` deploys the CloudFormation stack and calls method `register` to save outputs to `~/.autogluon/cloud.yaml`.

---

Follow Up PRs in order:

  - [ ] CLI equivalent (`autogluon-cloud bootstrap/status/teardown`) built on the same setup engine
  - [ ] Wire the config auto-load into `CloudPredictor.__init__` so users don't need to pass `cloud_output_path=` after `bootstrap()`
  - [ ] Update `docs/tutorials/autogluon-cloud.md` to lead with the `agc.initialize()` quick setup; add `init/status/teardown` to `docs/api.rst`

Note: Used opus 4.7 for development, please review carefully.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 
 from setuptools import setup
 
@@ -14,6 +15,25 @@ def create_version_file(*, version):
     with open(version_path, "w") as f:
         f.write(f'"""This is the {AUTOGLUON}.{CLOUD} version file."""\n')
         f.write("__version__ = '{}'\n".format(version))
+
+
+def sync_templates():
+    """Copy canonical CloudFormation templates into the package so they ship in the wheel.
+
+    The templates at repo-root `/cloudformation/` are the source of truth (referenced
+    from docs by public URL). At build time we mirror them under the `autogluon.cloud`
+    package so `importlib.resources` can locate them at runtime without depending on the
+    repo layout.
+    """
+    src_dir = "cloudformation"
+    dst_dir = os.path.join("src", AUTOGLUON, CLOUD, "templates")
+    if not os.path.isdir(src_dir):
+        # Building from an sdist that already has the copies — nothing to do.
+        return
+    os.makedirs(dst_dir, exist_ok=True)
+    for filename in os.listdir(src_dir):
+        if filename.endswith(".yaml"):
+            shutil.copyfile(os.path.join(src_dir, filename), os.path.join(dst_dir, filename))
 
 
 def update_version(version, use_file_if_exists=True, create_file=False):
@@ -64,7 +84,11 @@ def default_setup_args(*, version):
         include_package_data=True,
         python_requires=PYTHON_REQUIRES,
         package_data={
-            "autogluon.cloud": ["default_cluster_configs/*.yaml", "utils/autogluon_dlc.json"],
+            "autogluon.cloud": [
+                "default_cluster_configs/*.yaml",
+                "utils/autogluon_dlc.json",
+                "templates/*.yaml",
+            ],
         },
         classifiers=[
             "Development Status :: 4 - Beta",
@@ -136,6 +160,7 @@ extras_require["tests"] = test_requirements
 
 if __name__ == "__main__":
     create_version_file(version=version)
+    sync_templates()
     setup_args = default_setup_args(version=version)
     setup(
         install_requires=install_requires,
--- a/src/autogluon/cloud/backend/constant.py
+++ b/src/autogluon/cloud/backend/constant.py
@@ -5,3 +5,6 @@ TIMESERIES_SAGEMAKER = "timeseries_sagemaker"
 RAY = "ray"
 RAY_AWS = "ray_aws"
 TABULAR_RAY_AWS = "tabular_ray_aws"
+
+# Backends the top-level setup API can provision via CloudFormation.
+SUPPORTED_SETUP_BACKENDS = (SAGEMAKER, RAY_AWS)
--- a/src/autogluon/cloud/backend/constant.py
+++ b/src/autogluon/cloud/backend/constant.py
@@ -7,4 +7,4 @@ RAY_AWS = "ray_aws"
 TABULAR_RAY_AWS = "tabular_ray_aws"
 
 # Backends the top-level setup API can provision via CloudFormation.
-SUPPORTED_SETUP_BACKENDS = (SAGEMAKER, RAY_AWS)
+SUPPORTED_BACKENDS = (SAGEMAKER, RAY_AWS)
--- a/src/autogluon/cloud/cloud_setup.py
+++ b/src/autogluon/cloud/cloud_setup.py
@@ -2,162 +2,190 @@
 
 Usage::
 
-    import autogluon.cloud as agc
+    from autogluon.cloud import bootstrap, register, status, teardown
 
-    agc.initialize(backend="sagemaker", region="us-east-1")   # deploy CFN + save config
-    agc.initialize(role_arn=..., bucket=...)                  # save config, skip CFN
-    agc.status()                                             # dict of health checks
-    agc.teardown(delete_bucket_contents=True)                # delete CFN + config
+    bootstrap()                                          # deploy CFN + save config
+    register(role_arn=..., bucket=..., region=...)       # save existing resources
+    status()                                             # dict of health checks
+    teardown(delete_bucket_contents=True)                # delete CFN + config
 """
 
 from __future__ import annotations
 
-import typing
 from importlib import resources
 from typing import Any, Dict, Literal, Optional
 
 import boto3
 from botocore.exceptions import BotoCoreError, ClientError, NoCredentialsError
 
+from .backend.constant import RAY_AWS, SAGEMAKER, SUPPORTED_SETUP_BACKENDS
 from .config import (
-    Profile,
-    delete_profile,
+    CloudConfig,
+    delete_config,
     get_config_path,
     load_config,
-    upsert_profile,
+    save_config,
 )
 
-__all__ = ["initialize", "status", "teardown"]
+__all__ = ["bootstrap", "register", "status", "teardown"]
 
-Backend = Literal["sagemaker", "ray_aws"]
-_SUPPORTED_BACKENDS = typing.get_args(Backend)
+BackendName = Literal[SAGEMAKER, RAY_AWS]
 
 
-def initialize(
+def bootstrap(
     *,
-    backend: Backend = "sagemaker",
-    region: Optional[str] = None,
+    backend: BackendName = "sagemaker",
     stack_name: Optional[str] = None,
-    role_arn: Optional[str] = None,
-    bucket: Optional[str] = None,
-    aws_profile: Optional[str] = None,
-    profile_name: str = "default",
+    session: Optional[boto3.Session] = None,
 ) -> None:
-    """Provision AG-Cloud resources and save a config profile.
+    """Deploy the CloudFormation stack and persist resource identifiers.
 
-    Either deploy a CloudFormation stack (creates an IAM role + S3 bucket
-    from scratch) or persist pre-existing resources — pass ``role_arn`` and
-    ``bucket`` together to skip CloudFormation. The persisted profile is the
-    canonical record; call :func:`status` to inspect it.
+    On completion the IAM role and S3 bucket created by the stack are saved
+    to ``~/.autogluon/cloud.yaml`` via :func:`register`. If you
+    already have an IAM role and bucket in place, call :func:`register`
+    directly and skip this function entirely.
 
     Parameters
     ----------
     backend
         Which AG-Cloud backend to provision.
-    region
-        AWS region. Falls back to the boto3-configured region, then ``us-east-1``.
     stack_name
         CloudFormation stack name. Auto-generated as ``ag-cloud-<backend>``
         if not given.
-    role_arn, bucket
-        If BOTH are given, skip CloudFormation and record these pre-existing
-        resources.
-    aws_profile
-        Local AWS profile name (from ``~/.aws/credentials`` or SSO) to use as
-        the base identity. Defaults to the standard boto3 credential chain.
-    profile_name
-        Name of the AG-Cloud profile entry in ``~/.autogluon/cloud.yaml``.
+    session
+        A ``boto3.Session`` to use for AWS calls. If ``None``, a default
+        session is constructed from the standard credential chain
+        (env vars, ``~/.aws/credentials``, SSO, instance profile).
 
     Raises
     ------
     ValueError
-        If ``role_arn`` and ``bucket`` aren't both provided together, or if
-        an unknown backend is passed.
+        If an unknown backend is passed.
     RuntimeError
         If AWS credentials can't be detected.
     """
-    if (role_arn is None) ^ (bucket is None):
-        raise ValueError("`role_arn` and `bucket` must be provided together.")
-    if backend not in _SUPPORTED_BACKENDS:
-        raise ValueError(f"Unsupported backend {backend!r}. Choose from {_SUPPORTED_BACKENDS}.")
+    if backend not in SUPPORTED_SETUP_BACKENDS:
+        raise ValueError(f"Unsupported backend {backend!r}. Choose from {SUPPORTED_SETUP_BACKENDS}.")
 
-    session = _verified_session(aws_profile=aws_profile, region=region)
-    region = session.region_name or "us-east-1"
+    session, account = _verified_session(session)
+    region = session.region_name
+    if region is None:
+        raise RuntimeError(
+            "AWS region not configured. Set AWS_DEFAULT_REGION, run `aws configure`, "
+            "or pass `session=boto3.Session(region_name=...)`."
+        )
+    stack_name = stack_name or f"ag-cloud-{backend.replace('_', '-')}"
 
-    if role_arn is None:
-        stack_name = stack_name or f"ag-cloud-{backend.replace('_', '-')}"
-        print(f"Deploying CloudFormation stack {stack_name!r} (this typically takes ~1 minute)...")
-        role_arn, bucket = _provision_stack(session, stack_name=stack_name, backend=backend)
-        print(f"Stack {stack_name!r} deployed.")
+    print(f"Deploying CloudFormation stack {stack_name!r} (account {account}, region {region}, ~1 minute)...")
+    role_arn, bucket = _provision_stack(session, stack_name=stack_name, backend=backend)
+    print(f"Stack {stack_name!r} deployed.")
 
-    profile = Profile(
+    register(
+        role_arn=role_arn,
+        bucket=bucket,
+        region=region,
+        backend=backend,
+        stack_name=stack_name,
+    )
+
+
+def register(
+    *,
+    role_arn: str,
+    bucket: str,
+    region: str,
+    backend: BackendName = "sagemaker",
+    stack_name: Optional[str] = None,
+) -> None:
+    """Persist resource identifiers to ``~/.autogluon/cloud.yaml``.
+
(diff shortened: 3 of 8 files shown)


src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,284 @@
...
+    Profile,
+    delete_profile,
+    get_config_path,
+    load_config,
+    upsert_profile,
+)
+
+__all__ = ["initialize", "status", "teardown"]
+
+Backend = Literal["sagemaker", "ray_aws"]
```
shchur: This shadows the `Backend` class in `backend.py`. Let's just use `Literal["sagemaker"]`? Or put the list of supported backends to `backend/constant.py`?

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,284 @@
...
+def initialize(
+    *,
+    backend: Backend = "sagemaker",
+    region: Optional[str] = None,
+    stack_name: Optional[str] = None,
+    role_arn: Optional[str] = None,
+    bucket: Optional[str] = None,
+    aws_profile: Optional[str] = None,
+    profile_name: str = "default",
+) -> None:
```
shchur, line 53: I feel like `initialize` currently has two distinct roles, which makes the API confusing:
1. Deploy infrastructure
2. Register existing infrastructure

Depending on which of these 2 paths is taken, some arguments get ignored or have no effect.

For example, `role_arn`: is this the role that is assumed to execute CFN stack provisioning or the role that is saved to AG's profile? Also `profile_name` and `aws_profile` look like they are related, but actually refer to very different objects.

shchur: One solution is to split this into two methods: `initialize` (deploys CFN stack) and `register` (saves credentials to `~/.autogluon/cloud.yaml`).

setup.py
```diff
@@ -136,6 +160,7 @@ def default_setup_args(*, version):
 
 if __name__ == "__main__":
     create_version_file(version=version)
+    sync_templates()
```
shchur, line 163: Maybe we could just move these CFN templates to the desired location under `src/...` to avoid the need to copy stuff on setup?

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,284 @@
...
+
+def initialize(
+    *,
+    backend: Backend = "sagemaker",
+    region: Optional[str] = None,
+    stack_name: Optional[str] = None,
+    role_arn: Optional[str] = None,
+    bucket: Optional[str] = None,
+    aws_profile: Optional[str] = None,
+    profile_name: str = "default",
```
shchur: Do we really need multiple profiles here? I feel like for v1 just having a single cached config would make the implementation & API much simpler

shchur (commented):
Thanks! Left a few comments

setup.py
AnirudhDagar: as discussed online this would have been ideal, but we have old version of the package and docs referencing these links in master.

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: thanks, simplified and split the method based on the review suggestions now:
`bootstrap` & `register`

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: removed

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: good catch, i will define this in the `backend/constant.py` and import here.

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+    CloudConfig,
+    delete_config,
+    get_config_path,
+    load_config,
+    save_config,
+)
+
+__all__ = ["bootstrap", "register", "status", "teardown"]
+
+BackendName = Literal[SAGEMAKER, RAY_AWS]
```
shchur: This looks like a syntax error - I think `Literal` requires the arguments to be strings, not variables.

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+        A ``boto3.Session`` to use for AWS calls. If ``None``, a default
+        session is constructed from the standard credential chain
+        (env vars, ``~/.aws/credentials``, SSO, instance profile).
+
+    Raises
+    ------
+    ValueError
+        If an unknown backend is passed.
+    RuntimeError
+        If AWS credentials can't be detected.
```
shchur: Minor: I think we can skip the Raises block in general

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+    """
+    if backend not in SUPPORTED_SETUP_BACKENDS:
+        raise ValueError(f"Unsupported backend {backend!r}. Choose from {SUPPORTED_SETUP_BACKENDS}.")
```
shchur: Nice, TIL about the `!r` option

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+    session, account = _verified_session(session)
+    region = session.region_name
+    if region is None:
+        raise RuntimeError(
+            "AWS region not configured. Set AWS_DEFAULT_REGION, run `aws configure`, "
+            "or pass `session=boto3.Session(region_name=...)`."
+        )
+    stack_name = stack_name or f"ag-cloud-{backend.replace('_', '-')}"
+
+    print(f"Deploying CloudFormation stack {stack_name!r} (account {account}, region {region}, ~1 minute)...")
```
shchur, line 85: Should we maybe use `logging` instead of `print` to give users an option to suppress the outputs?

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+    to ``~/.autogluon/cloud.yaml`` via :func:`register`. If you
+    already have an IAM role and bucket in place, call :func:`register`
+    directly and skip this function entirely.
+
+    Parameters
+    ----------
+    backend
+        Which AG-Cloud backend to provision.
+    stack_name
+        CloudFormation stack name. Auto-generated as ``ag-cloud-<backend>``
```
shchur: Minor: The project uses `line-length=119` but currently the docstrings seem to use `line-length=89`

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+    ----------
+    role_arn
+        ARN of an IAM role suitable for SageMaker / Ray to assume.
+    bucket
+        S3 bucket name where AG-Cloud will read/write artifacts.
+    region
+        AWS region for AG-Cloud operations.
+    backend
+        Which AG-Cloud backend the resources are intended for.
+    stack_name
```
shchur, line 125: Do I understand correctly that this is only for our internal usage (to keep track of the CFN stack name)? Most users won't set this, right?

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+    except ClientError as e:
+        return e.response["Error"]["Message"]
+
+
+def _check_role(session: boto3.Session, role_arn: str) -> str:
+    """Verify the IAM role exists via iam:GetRole. Doesn't call sts:AssumeRole —
+    we only check existence, not the caller's permission to assume it.
+    """
+    try:
+        session.client("iam").get_role(RoleName=role_arn.rsplit("/", 1)[-1])
```
shchur: Roles might contain multiple slashes, should we do this instead?
```python
role_arn.split(":role/", 1)[-1]
```

src/autogluon/cloud/config.py
```diff
@@ -0,0 +1,72 @@
...
+        return Path(override).expanduser()
+    return Path.home() / ".autogluon"
+
+
+def get_config_path() -> Path:
+    return get_config_dir() / "cloud.yaml"
+
+
+@dataclass
+class CloudConfig:
```
shchur: Major: Currently the `CloudConfig` and the corresponding file can only hold a single `backend`. Should we instead structure the config file as `{backend}: {config}`? Something like
```yaml
sagemaker:
   role: ...
   bucket: ...
ray_aws:
   role: ...
```
Then we will retrieve the correct config based on the backend that the user specified in `CloudPredictor` or `FoundationModel`.

shchur: At first we will only have the `SageMaker` backend as the properly supported version, but I wonder if it makes sense to think about what happens when there are >1 backends present.

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+        "config_path": str(get_config_path()),
+        "config": config,
+        "checks": checks,
+    }
+
+
+def teardown(
+    *,
+    session: Optional[boto3.Session] = None,
+    delete_bucket_contents: bool = False,
```
shchur: I wonder if we can put any more guardrails in place, this looks like a really dangerous operation 😬 

Two ideas:
1. Maybe we can just tell the user to empty the bucket themselves? 
2. Ask to put in the bucket name as a confirmation.

I am leaning towards 1

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+        checks["stack"] = _check_stack(session, config.stack_name)
+    if check_role:
+        checks["role"] = _check_role(session, config.role_arn)
+
+    return {
+        "found": True,
+        "config_path": str(get_config_path()),
+        "config": config,
+        "checks": checks,
+    }
```
shchur: Two comments
1.   `status()` returns a loose `Dict[str, Any]` with two shapes (`found=True/False`) — let's maybe return a `TypedDict` or `dataclass`, or `None` if there is no config found.

2.  Drop `check_role` param — instead, handle `AccessDenied` gracefully in `_check_role`/`_check_stack` by returning something like `"ok (unverified)"` rather than `"failed"`. Currently they report failure when it's actually a caller permissions issue, not a broken resource.

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+        bucket=bucket,
+        region=region,
+        backend=backend,
+        stack_name=stack_name,
+    )
+
+
+def register(
+    *,
+    role_arn: str,
```
shchur: Let's use `role` instead of `role_arn` here for consistency with the SageMaker SDK naming

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,307 @@
...
+            StackName=stack_name,
+            TemplateBody=template.read_text(),
+            Capabilities=["CAPABILITY_NAMED_IAM"],
+        )
+    except ClientError as e:
+        if e.response["Error"]["Code"] != "AlreadyExistsException":
+            raise
+        print(f"Stack {stack_name!r} already exists — reusing it.")
+
+    cfn.get_waiter("stack_create_complete").wait(StackName=stack_name)
```
shchur: Comment from our good friend:

> `_provision_stack` waits on `stack_create_complete` even when the stack already exists — if it's already CREATE_COMPLETE, the waiter succeeds immediately, but if it's in UPDATE_ROLLBACK_COMPLETE or another terminal state, this could hang or error confusingly. Should either use `describe_stacks` to check current state after `AlreadyExistsException`, or just report the outputs directly.

src/autogluon/cloud/config.py
```diff
@@ -0,0 +1,72 @@
...
+
+from __future__ import annotations
+
+import os
+import stat
+from dataclasses import asdict, dataclass
+from pathlib import Path
+from typing import Optional
+
+import yaml
```
shchur, line 31: Let's put `pyyaml` into the requirements - now we rely on it as a transitive dependency from `autogluon.common` (not idea)

shchur (commented):
Just a bunch of minor comments, but overall this looks great!

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: yes, that's correct!

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: i see, wasn't aware of that. changing this now

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: iam:GetRole's `RoleName` parameter takes the bare role name without the path. Per the [IAM docs](https://docs.aws.amazon.com/boto3/latest/reference/services/iam/client/get_role.html), an ARN like `arn:aws:iam::123:role/service-role/MyRole` has path = `/service-role/` and role name = `MyRole`. And `RoleName` only accepts the bare name (`MyRole`), it rejects path-prefixed values.
`rsplit("/", 1)[-1]` always gets the last segment (the role name) regardless of how many path components exist in between, so it works.

I ran a quick smoke test against real IAM to confirm:
```python
iam.get_role(RoleName="NonExistentBareName")
# → NoSuchEntity (format accepted, role just doesn't exist)

iam.get_role(RoleName="service-role/NonExistentName")
# → ValidationError: roleName must contain only alphanumeric and +=,.@_-

iam.get_role(RoleName="team/prod/NonExistentName")
# → ValidationError: same
```

src/autogluon/cloud/config.py
AnirudhDagar: https://github.com/autogluon/autogluon-cloud/blob/83853b1600869d3d2f83e200e0157085c5e1bee0/setup.py#L116

pyyaml is already present

src/autogluon/cloud/config.py
AnirudhDagar: good catch, even though most people will probably use only one backend but we can't be limiting this.

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: yeah, i removed the s3 bucket deletion then. If it is empty it will anyway be removed with the stack removal. If it is not empty, let's not touch it.

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: Thanks that makes sense, created a dataclass

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: I had logging INFO level before but it is by default set to the WARNING level, we could switch the default to INFO and replace prints by a logger. I'd tackle that in a separate PR if that's ok.

Because the prints are supressed, it might feel like the process hangs after calling bootstrap as it takes atleast 30-45seconds to deploy the stack.

src/autogluon/cloud/config.py
AnirudhDagar: in that case teardown will also accept backend, in case a user wants to specifically get rid of one backend stack

AnirudhDagar (commented):
Thanks for the review @shchur, I've addressed the comments and pushed an update for the same.

src/autogluon/cloud/config.py
shchur: My bad

src/autogluon/cloud/backend/constant.py
```diff
@@ -5,3 +5,6 @@
 RAY = "ray"
 RAY_AWS = "ray_aws"
 TABULAR_RAY_AWS = "tabular_ray_aws"
+
+# Backends the top-level setup API can provision via CloudFormation.
+SUPPORTED_SETUP_BACKENDS = (SAGEMAKER, RAY_AWS)
```
shchur: Minor: Why not just `SUPPORTED_BACKENDS`?

src/autogluon/cloud/config.py
```diff
@@ -0,0 +1,96 @@
...
+
+import os
+import stat
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import Dict, Optional
+
+import yaml
+
+CONFIG_DIR_ENV = "AUTOGLUON_CLOUD_CONFIG_DIR"
```
shchur: We typically use the `AG_*` prefix for env variables. How about just `AG_CONFIG_DIR` since `~/.autogluon` isn't cloud-specific?

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    config = load_config()
+    if config is None:
+        return {}
+
+    reports: Dict[str, StatusReport] = {}
+    for name in list(config.backends):
+        backend_config = config.backends.get(name)
+        if backend_config is None:
+            continue
```
shchur: Looks redundant, what about just
```python
for name, backend_config in config.backends.items():
```

src/autogluon/cloud/cloud_setup.py
```diff
@@ -0,0 +1,335 @@
...
+            f"Delete it via the CloudFormation console and re-run."
+        )
+    return outputs["RoleARN"], outputs["BucketName"]
+
+
+# AWS error codes that mean "the caller lacks IAM permission to read this", as
+# distinct from "the resource doesn't exist". We surface these as ``"unverified"``
+# rather than ``"failed"`` so users don't think their setup is broken when really
+# it's just a permissions gap on the side of whoever is running ``status()``.
+_PERMISSION_ERROR_CODES = frozenset({"AccessDenied", "AccessDeniedException", "Forbidden", "UnauthorizedOperation"})
```
shchur: Let's just put it inside the `_is_permission_error` method - variable not used anywhere else

setup.py
```diff
@@ -16,6 +17,25 @@ def create_version_file(*, version):
         f.write("__version__ = '{}'\n".format(version))
 
 
+def sync_templates():
```
shchur, line 20: Let's just duplicate the templates under `src/autogluon/cloud/templates/` in this PR instead of having this copy-at-runtime behavior with `sync_templates`. 

We can then remove the templates from the project root once we update the docs in a follow-up PR.

shchur (approved):
Thanks! Feel free to merge after addressing the remaining comments

src/autogluon/cloud/config.py
AnirudhDagar: that is true, updating to AG_CONFIG_DIR

setup.py
AnirudhDagar: Discussed offline, we'll tackle this later

src/autogluon/cloud/cloud_setup.py
AnirudhDagar: sorry i missed this, after making some changes now it looks redundant. will switch to the one liner

AnirudhDagar:
Thanks @shchur for multiple rounds of reviews and helping make the design much better! I'll merge once the CI is green.

PR #214, opened 2026-05-11
Add fit_predict single-job flow and integration tests

melopeo:
Issue #, if available

Description of changes:

Adds `fit_predict` to `TimeSeriesCloudPredictor` so users can run fit and
predict in a single SageMaker job, avoiding the ~6-minute startup cost of a
second job. This is the pattern we want for foundation-model forecasting
(Chronos-2, Chronos, Chronos-Bolt, etc.) where "fit" just loads pretrained
weights.

### Usage

```python
from autogluon.cloud import TimeSeriesCloudPredictor

predictor = TimeSeriesCloudPredictor(cloud_output_path="s3://...")

predictions = predictor.fit_predict(
    train_data= train_df,
    predictor_init_args=dict(
        target="target",
        prediction_length=5,
    ),
    predictor_fit_args=dict(
        hyperparameters={"Chronos2": {"model_path": "autogluon/chronos-2-small"}},
    ),
)
```

With known covariates:
```python
predictions = predictor.fit_predict(
    train_data=train_df,
    known_covariates=future_covariates_df,
    predictor_init_args=dict(
        target="target",
        prediction_length=5,
        known_covariates_names=["cov1", "cov2"],
    ),
    predictor_fit_args=dict(
        hyperparameters={"Chronos2": {"model_path": "autogluon/chronos-2-small"}},
    ),
)
```


src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +316,126 @@ def predict_proba(
...
+        predictor_metadata = {
+            "id_column": self.id_column,
+            "timestamp_column": self.timestamp_column,
+            "target_column": self.target_column,
+        }
+        backend_kwargs.setdefault("autogluon_sagemaker_estimator_kwargs", {}).setdefault("hyperparameters", {})[
+            "predictor_metadata"
+        ] = json.dumps(predictor_metadata)
+
+        backend_kwargs = self.backend.parse_backend_fit_kwargs(backend_kwargs)
```
shchur: This logic seems to be mostly copied from `fit()`. 

Could we rather call `self.fit(...)` instead of copying code + `self.backend.fit(...)`?

We can then inject the `fit_predict` kwarg to the `backend_kwargs` using `_construct_ag_args` or something similar to avoid extending the backend API even further with a new kwarg.

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +316,126 @@ def predict_proba(
     ) -> Optional[pd.DataFrame]:
         raise ValueError(f"{self.__class__.__name__} does not support predict_proba operation.")
 
+    def fit_predict(
+        self,
+        test_data: Optional[Union[str, pd.DataFrame]] = None,
```
shchur: How about we change the signature as follows
```python
train_data: Union[str, pd.DataFrame],
known_covariates: Optional[Union[str, pd.DataFrame]] = None,
```
and then insert `train_data` + `known_covariates` into the kwargs passed to `self.fit`?

The concept of `test_data` is a bit weird in context of time series (is it historical data or known future covariates?), also we need to pass known covariates.

Currently `TSCPredictor.fit(...)` does not take `train_data` as an explicit kwarg (rather expects it inside `predictor_fit_kwargs`), but I think it makes sense to make the data explicit arguments for better readability.

src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -168,6 +168,21 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
...
 
+    # fit_predict: run prediction inside the same training job and persist results so
+    # the caller can download them alongside the model artifacts. This avoids paying
+    # the SageMaker startup cost twice.
+    if ag_args.get("fit_predict", False):
+        print("Running in-job prediction for fit_predict")
+        if predictor_type == "timeseries":
+            predictions = predictor.predict(training_data)
+            predictions = pd.DataFrame(predictions).reset_index()
+        else:
```
shchur: Let's explicitly handle `tabular` / `multimodal` cases. In both of them, I don't think we want to make predictions on `train_data` - would make more sense to accept `test_data: pd.DataFrame | str` in `fit_predict(...)` and make predictions on this new data.

It's okay to deal with that later (or maybe even skip mulitmodal support completely), but at least now predicting on train data seems not ideal for tabular.

src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -168,6 +168,21 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
             lb = predictor.leaderboard(silent=False)
             lb.to_csv(f"{args.output_data_dir}/leaderboard.csv")
 
+    # fit_predict: run prediction inside the same training job and persist results so
+    # the caller can download them alongside the model artifacts. This avoids paying
+    # the SageMaker startup cost twice.
+    if ag_args.get("fit_predict", False):
+        print("Running in-job prediction for fit_predict")
+        if predictor_type == "timeseries":
+            predictions = predictor.predict(training_data)
```
shchur: We also need to handle the known covariates passed from fit_predict (see other comment)

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +316,126 @@ def predict_proba(
...
+        return pd.read_csv(predictions_path)
+
+    def download_fit_predict_results(self, save_path: Optional[str] = None) -> str:
+        """
+        Download the predictions produced by a completed ``fit_predict`` job.
+
+        Returns the path to ``predictions.csv`` extracted from the job's
+        ``output.tar.gz``.
+        """
+        return self.backend.download_fit_predict_results(save_path=save_path)
```
shchur: Currently `self.backend` is of type `Backend` that does not define the `download_fit_predict_results` method. Also the `download_*` part seems closely coupled to the SageMaker backend - another backend like Lambda might not require any downloading (the predictions would just be returned as a payload from the invocation).

One idea is to add an abstract method `get_fit_predict_results(self) -> pd.DataFrame` to `Backend` with a concrete implementation in `SageMakerBackend` that downloads the results from the training job. 

We can still keep the option to save to a concrete path with smth like
```python
def get_fit_predict_results(self, save_path: Optional[str] = None) -> pd.DataFrame:
    predictions = self.backend.get_fit_predict_results()
    if save_path:
        predictions.to_csv(save_path, index=False)
    return predictions
```

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +316,126 @@ def predict_proba(
...
+            timestamp_column=timestamp_column,
+            static_features=static_features,
+            framework_version=framework_version,
+            job_name=job_name,
+            instance_type=instance_type,
+            instance_count=instance_count,
+            volume_size=volume_size,
+            custom_image_uri=custom_image_uri,
+            wait=wait,
+            fit_predict=True,
```
shchur: Minor: What about something more distinct from the method name like `predict_after_fit`?

shchur (commented):
Thanks Pedro! Left a few comments

src/autogluon/cloud/backend/backend.py
```diff
@@ -172,3 +172,14 @@ def predict(self, test_data: Union[str, pd.DataFrame], **kwargs) -> Union[pd.Dat
...
         raise NotImplementedError
+
+    def get_fit_predict_results(self) -> pd.DataFrame:
+        """
+        Retrieve predictions produced by a completed ``fit_predict`` job.
+
+        Implementations decide how to materialize the predictions. For SageMaker
+        this downloads ``predictions.csv`` from the training job's output
+        artifact; a future Lambda/Batch backend may return the invocation
+        payload directly.
```
shchur: This comment is all about implementation that are unrelated to the base class, do we need it here?

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -950,6 +961,37 @@ def download_predict_results(self, job_name: Optional[str] = None, save_path: Op
 
         return results_save_path
 
+    def get_fit_predict_results(self) -> pd.DataFrame:
```
shchur, line 967: It looks like there is a lot of logic duplication with `download_predict_results` right above. Can both methods somehow rely on the same logic? Or maybe one method can use the other?

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -950,6 +961,37 @@ def download_predict_results(self, job_name: Optional[str] = None, save_path: Op
...
+        job_name = self._fit_job.job_name
+        assert job_name is not None, "No fit job found. Call `fit_predict()` first."
+        output_tarball = self.cloud_output_path + f"/model/{job_name}/output/output.tar.gz"
+        assert is_s3_url(output_tarball)
+        bucket, key = s3_path_to_bucket_prefix(output_tarball)
+
+        with tempfile.TemporaryDirectory(prefix="ag_fit_predict_") as tmpdir:
+            tarball_local = os.path.join(tmpdir, "output.tar.gz")
+            logger.log(20, f"Downloading fit_predict output from s3://{bucket}/{key}")
+            s3 = boto3.client("s3")
```
shchur: This doesn't respect the AWS config (profile, role, etc). Looks like we are not using boto3 anywhere else in the SageMakerBackend - I would recommend using patterns like `self.sagemaker_session.download_data` as in other methods.

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -950,6 +961,37 @@ def download_predict_results(self, job_name: Optional[str] = None, save_path: Op
...
+            s3.download_file(bucket, key, tarball_local)
+            with tarfile.open(tarball_local) as tf:
+                tf.extractall(tmpdir)
```
shchur: This is unsafe https://codeql.github.com/codeql-query-help/python/py-tarslip/, one possible solution to set `filter='data'` (https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall)

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -212,6 +218,8 @@ def fit(
         wait: bool = True,
         autogluon_sagemaker_estimator_kwargs: Optional[Dict] = None,
         fit_kwargs: Optional[Dict] = None,
+        ag_args_extras: Optional[Dict[str, Any]] = None,
+        known_covariates: Optional[pd.DataFrame] = None,
```
shchur, line 221: Should we also support s3 paths to `known_covariates: str | pd.DataFrame | None` (same as for train data)?

src/autogluon/cloud/backend/timeseries_sagemaker_backend.py
```diff
@@ -124,6 +136,31 @@ def fit(
...
+            cols = known_covariates_df.columns.to_list()
+            for required in (id_column, timestamp_column):
+                if required not in cols:
+                    raise ValueError(f"`known_covariates` must contain column '{required}'.")
+            # Reorder so id and timestamp come first (parity with _preprocess_data).
+            timestamp_index = cols.index(timestamp_column)
+            cols.insert(0, cols.pop(timestamp_index))
+            id_index = cols.index(id_column)
+            cols.insert(0, cols.pop(id_index))
+            known_covariates_df = known_covariates_df[cols]
```
shchur: I'm a bit confused, why do we need this block? How do we package `known_covariates` into the data that goes to the predictor? I see that `known_covariates` was added as a pass-through argument for a few methods but I'm struggling a bit with the big picture. 

Could we instead package `known_covariates` into `predictor_fit_args` (same as how we handle `train_data` and `tuning_data`) and then just pop `known_covariates` out of the `predictor_fit_args` inside the training job before training? This would also allow us to avoid e.g. introducing the new overload for `parse_backend_fit_kwargs` and extending the signatures of other methods.

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +317,139 @@ def predict_proba(
...
+            Name of the item ID column.
+        timestamp_column: str, default = "timestamp"
+            Name of the timestamp column.
+        static_features: Optional[pd.DataFrame]
+            Optional metadata attributes per item.
+        framework_version, job_name, instance_type, instance_count, volume_size,
+        custom_image_uri, wait, backend_kwargs:
+            Same semantics as ``fit()``.
+        save_path: Optional[str]
+            Local directory to save the downloaded predictions. Defaults to
```
shchur: Minor: The docstrings are formatted with `line-length = 90`, which is inconsistent with the rest of the codebase that uses `line-length = 119` (see `pyproject.toml`)

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +317,139 @@ def predict_proba(
...
+            raise ValueError(
+                "`train_data` must be passed as an explicit argument to `fit_predict`, not via `predictor_fit_args`."
+            )
+        predictor_fit_args["train_data"] = train_data
+
+        if backend_kwargs is None:
+            backend_kwargs = {}
+        else:
+            backend_kwargs = dict(backend_kwargs)
+        backend_kwargs.setdefault("ag_args_extras", {})["predict_after_fit"] = True
```
shchur: I am confused about the role of the `ag_args_extras` and why it lives inside `backend_kwargs`. Can you please explain the reasoning for putting it inside `backend_kwargs` and how `ag_args_extras` is different from existing parameters like `ag_args`?

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +317,139 @@ def predict_proba(
...
+        framework_version: str = "latest",
+        job_name: Optional[str] = None,
+        instance_type: str = "ml.m5.2xlarge",
+        instance_count: int = 1,
+        volume_size: int = 100,
+        custom_image_uri: Optional[str] = None,
+        wait: bool = True,
+        save_path: Optional[str] = None,
+        backend_kwargs: Optional[Dict] = None,
+    ) -> Optional[pd.DataFrame]:
```
shchur, line 341: Related to the example in the PR description: should we do some validation on `known_covariates_names` in `predictor_init_args` and the `known_covariates` passed to this method? I can imagine multiple scenarios where the user misconfigured something (e.g. forgot to put names in init args) that will only be caught after the training job is complete - which would be really frustrating for the user

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +317,139 @@ def predict_proba(
...
+        """
+        Fit and predict in a single SageMaker training job.
+
+        This is useful for foundation-model forecasting workflows (e.g. Chronos-2)
+        where "fit" is essentially loading a pretrained model. Running fit and
+        predict in the same job avoids paying the SageMaker startup cost twice.
```
shchur: Nit: "paying the cost" sounds like it's about the money, even though the main motivation is time, as I understand.

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,51 @@ def test_timeseries(test_helper, framework_version):
...
+            train_data=train_data,
+            predictor_init_args=dict(target="target", prediction_length=3),
+            predictor_fit_args=dict(
+                hyperparameters=hyperparameters,
+            ),
+            framework_version=framework_version,
+            custom_image_uri=training_custom_image_uri,
+        )
+
+        assert isinstance(predictions, pd.DataFrame)
```
shchur: Let's check that predictions contains the expected item_ids, columns, and number of values per item?

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,51 @@ def test_timeseries(test_helper, framework_version):
...
+# each.
+@pytest.mark.parametrize(
+    "model_name, hyperparameters",
+    [
+        ("chronos", {"Chronos": {"model_path": "tiny"}}),
+        ("chronos_bolt", {"Chronos": {"model_path": "bolt_small"}}),
+        ("chronos2", {"Chronos2": {"model_path": "autogluon/chronos-2-small"}}),
+    ],
+)
+def test_timeseries_fit_predict_chronos(test_helper, framework_version, model_name, hyperparameters):
```
shchur: Can we also test that fit_predict works with known_covariates?

src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -168,6 +174,35 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
             lb = predictor.leaderboard(silent=False)
             lb.to_csv(f"{args.output_data_dir}/leaderboard.csv")
 
+    # fit_predict: run prediction inside the same training job and persist results so
+    # the caller can download them alongside the model artifacts. This avoids paying
+    # the SageMaker startup cost twice.
+    #
+    # Currently scoped to the timeseries predictor, which is the only one that
+    # exposes `fit_predict` from the client side. Tabular / multimodal support
+    # (with explicit `test_data` plumbing) can land in a follow-up.
```
shchur: This looks like an LLM comment that shouldn't be commented to the codebase

src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -168,6 +174,35 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
...
+        predict_kwargs = {}
+        if args.known_covariates:
+            kc_file = get_input_path(args.known_covariates)
+            kc_df = pd.read_csv(kc_file)
+            # Column layout set by TimeSeriesSagemakerBackend: id, timestamp, cov1, ..., covN
+            id_col = kc_df.columns[0]
+            ts_col = kc_df.columns[1]
+            kc_df[ts_col] = pd.to_datetime(kc_df[ts_col])
+            known_covariates = TimeSeriesDataFrame.from_data_frame(kc_df, id_column=id_col, timestamp_column=ts_col)
+            predict_kwargs["known_covariates"] = known_covariates
```
shchur: Code looks mostly duplicated from `prepare_timeseries_dataframe`. Can we just extend the method (e.g. make `target` & `static_features` handling optional) and reuse it here instead of copying the logic?

Alternatively, we could just clean up the logic for handling TSDF - just rename the `id_column` to `item_id` and `timestamp_column` to `timestamp` locally before sending the data to SageMaker.

src/autogluon/cloud/backend/timeseries_sagemaker_backend.py
```diff
@@ -43,6 +57,25 @@ def _preprocess_data(
...
+        Thin wrapper over ``_preprocess_data`` with ``target=None`` and no static features, so the frame keeps
+        layout ``[id, timestamp, cov1, ..., covN]``.
+        """
+        return self._preprocess_data(
+            data=data,
+            id_column=id_column,
+            timestamp_column=timestamp_column,
+            target=None,
+            static_features=None,
+        )
```
shchur: This method seems redundant, why cannot we just replace its call with the call to `self._preprocess_data`?

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -950,6 +958,37 @@ def download_predict_results(self, job_name: Optional[str] = None, save_path: Op
...
+                csv_member = next(
+                    (m for m in tf.getmembers() if m.isfile() and os.path.basename(m.name) == "predictions.csv"),
+                    None,
+                )
+                assert csv_member is not None, (
+                    f"Could not find predictions.csv in {tarball_local}. "
+                    "Did the training job run with `fit_predict=True`?"
+                )
+                with tf.extractfile(csv_member) as fp:
+                    return pd.read_csv(fp)
```
shchur: We already know under which path we saved the predictions, couldn't we just directly attempt to extract this file with something like the following
```python
  with tarfile.open(tarball_local) as tf:
      try:
          fp = tf.extractfile("predictions.csv")
      except KeyError:
          raise RuntimeError(
              f"predictions.csv not found in {tarball_local}. "
              "Did the training job run with `fit_predict=True`?"
          )
      return pd.read_csv(fp)
```

src/autogluon/cloud/backend/timeseries_sagemaker_backend.py
```diff
@@ -124,6 +159,18 @@ def fit(
             )
         predictor_fit_args["train_data"] = train_data
         predictor_fit_args["tuning_data"] = tuning_data
+
+        if known_covariates is not None:
+            if not predict_after_fit:
+                raise ValueError(
+                    "`known_covariates` is only meaningful for `fit_predict`; use `predict()` for separate inference."
```
shchur: The `backend` has no knowledge of the `fit_predict` method. A better phrasing would be smth like `known_covariates should only provided if predict_after_fit=True`

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +319,202 @@ def predict_proba(
...
+            The historical time-series data to train on and forecast from. Either a pandas DataFrame or a local / S3
+            path to a CSV.
+        predictor_init_args: dict
+            Init args for the predictor (must include ``prediction_length``).
+        predictor_fit_args: Optional[dict], default = None
+            Additional fit args for the predictor. Must not contain a ``train_data`` key — pass ``train_data`` as the
+            explicit argument above.
+        known_covariates: Optional[Union[str, pd.DataFrame]], default = None
+            Values of the known covariates for each time series during the forecast horizon. Either a pandas
+            DataFrame or a local / S3 path to a CSV. Forwarded to ``TimeSeriesPredictor.predict`` in the container.
```
shchur: Only CSV? I thought that e.g. parquet was also supported

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +319,202 @@ def predict_proba(
...
+
+        Returns
+        -------
+        pd.DataFrame
+            Predictions for the forecast horizon.
+        """
+        predictions = self.backend.get_fit_predict_results()
+        if save_path is not None:
+            save_path = os.path.abspath(os.path.expanduser(save_path))
+            os.makedirs(os.path.dirname(save_path) or ".", exist_ok=True)
```
shchur: Why do we need `or "."` here?

src/autogluon/cloud/predictor/timeseries_cloud_predictor.py
```diff
@@ -316,6 +319,202 @@ def predict_proba(
...
+        return known_covariates
+
+    def get_fit_predict_results(self, save_path: Optional[str] = None) -> pd.DataFrame:
+        """
+        Retrieve predictions produced by a completed ``fit_predict`` job.
+
+        Parameters
+        ----------
+        save_path: Optional[str], default = None
+            If set, the predictions are additionally written to this path as a CSV (directories are created as
```
shchur: It's not 100% clear that this should be a path to a directory and that the predictions are saved under `Path(save_path) / "predictions.csv"`

src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -156,6 +167,11 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
         image_column = ag_args["image_column"]
         tuning_data[image_column] = tuning_data[image_column].apply(lambda path: os.path.join(tune_images_dir, path))
 
+    # `known_covariates` rides through predictor_fit_args for the in-job predict
+    # step of `fit_predict`. TimeSeriesPredictor.fit doesn't accept it, so pop
```
shchur: In the previous version of this PR the `known_covariates` were passed via a separate input channel to the SageMaker training job. What is the reasoning for this change? Are there some limits on the overall serialized size of the `predictor_fit_args`? (i.e. is there a max size of the known covariates DF that will exceed some sagemaker limits?)

src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -168,6 +184,31 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
...
+                f"`fit_predict` is only supported for predictor_type='timeseries', got '{predictor_type}'."
+            )
+        print("Running in-job prediction for fit_predict")
+        predict_kwargs = {}
+        if known_covariates_df is not None:
+            predict_kwargs["known_covariates"] = prepare_timeseries_dataframe(known_covariates_df, target=None)
+        predictions = predictor.predict(training_data, **predict_kwargs)
+        predictions = pd.DataFrame(predictions).reset_index()
+        # AutoGluon's TimeSeriesDataFrame normalizes the id column to "item_id". Restore the user's original
+        # id column name so the predictions DataFrame matches the input schema.
```
shchur: Do real-time & batch prediction endpoints currently rename `"item_id"` to the original `id_column`? I thought that currently `id_column` was only passed to the constructor for `TimeSeriesDataFrame` and the predictions were always returned in the original AutoGluon format, i.e. with the `"item_id"` and `"timestamp"` column names.

The behavior of this method should stay consistent with the other methods (real-time and batch prediction) imo.

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,118 @@ def test_timeseries(test_helper, framework_version):
             ),
             predict_real_time_kwargs=dict(static_features=static_features),
         )
+
+
+# Chronos (T5-based), Chronos-Bolt, and Chronos-2 are all exposed through the TimeSeries predictor's hyperparameters
+# dict under slightly different keys; the parametrization below exercises the ``fit_predict`` plumbing for each.
+# The ``with_covariates`` flag opts a case into the ``known_covariates`` code path (Chronos-2 only — covariates
+# support is most relevant there). When set, the test pulls a public retail-sales fixture from the AutoGluon S3
+# bucket; otherwise it uses the small CI csv fixture.
```
shchur: Again, this looks like an LLM-generated comment that shouldn't have leaked into the codebase. I would only keep comments near the places in the code where something unexpected / non-obvious happens.

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,118 @@ def test_timeseries(test_helper, framework_version):
...
+    target = "Sales"
+    id_column = "id"
+    timestamp_column = "timestamp"
+    train_df = pd.read_parquet("https://autogluon.s3.amazonaws.com/datasets/timeseries/retail_sales/train.parquet")
+    train_df[timestamp_column] = pd.to_datetime(train_df[timestamp_column])
+    test_df = pd.read_parquet("https://autogluon.s3.amazonaws.com/datasets/timeseries/retail_sales/test.parquet")
+    test_df[timestamp_column] = pd.to_datetime(test_df[timestamp_column])
+    known_covariates_df = test_df.drop(columns=target)
+    train_df[id_column] = pd.to_numeric(train_df[id_column])
+    known_covariates_df[id_column] = pd.to_numeric(known_covariates_df[id_column])
```
shchur: Why are we casting the id_column to numeric dtype? Object/string dtype is okay for item_ids.

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,118 @@ def test_timeseries(test_helper, framework_version):
...
+    of ``train_df`` minus the target column, so it carries the future values of the known covariates over the
+    forecast horizon.
+    """
+    target = "Sales"
+    id_column = "id"
+    timestamp_column = "timestamp"
+    train_df = pd.read_parquet("https://autogluon.s3.amazonaws.com/datasets/timeseries/retail_sales/train.parquet")
+    train_df[timestamp_column] = pd.to_datetime(train_df[timestamp_column])
+    test_df = pd.read_parquet("https://autogluon.s3.amazonaws.com/datasets/timeseries/retail_sales/test.parquet")
+    test_df[timestamp_column] = pd.to_datetime(test_df[timestamp_column])
```
shchur, line 68: Nit: Is this necessary? I thought parquet already stores the correct dtype

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,118 @@ def test_timeseries(test_helper, framework_version):
...
+        )
+    else:
+        target = "target"
+        id_column = "item_id"
+        timestamp_column = "timestamp"
+        prediction_length = 3
+        train_data_csv = "timeseries_train.csv"
+        train_data = train_data_csv
+        known_covariates = None
+        predictor_init_args = dict(target=target, prediction_length=prediction_length)
```
shchur: Minor: Do we need these two full blocks for each branch? Why not just have something like
```python
train_data, known_covariates, known_covariates_names = _load_retail_sales_with_covariates()
if not with_covariates:
    known_covariates = None
    known_covariates_names = None
```

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,118 @@ def test_timeseries(test_helper, framework_version):
...
+        ("chronos2_with_covs", {"Chronos2": {"model_path": "autogluon/chronos-2-small"}}, True),
+    ],
+)
+def test_timeseries_fit_predict_chronos(test_helper, framework_version, model_name, hyperparameters, with_covariates):
+    timestamp = test_helper.get_utc_timestamp_now()
+    if with_covariates:
+        target = "Sales"
+        id_column = "id"
+        timestamp_column = "timestamp"
+        prediction_length = 13
```
shchur: Minor: Looks weird that we are setting these constants outside, even though they are very closely coupled to the internals of the `_load_retail_sales_with_covariates` method. Why not, e.g., return `predictor_init_args` as a dict from `_load_retail_sales_with_covariates`?

tests/unittests/timeseries/test_timeseries.py
```diff
@@ -51,3 +54,118 @@ def test_timeseries(test_helper, framework_version):
...
+# dict under slightly different keys; the parametrization below exercises the ``fit_predict`` plumbing for each.
+# The ``with_covariates`` flag opts a case into the ``known_covariates`` code path (Chronos-2 only — covariates
+# support is most relevant there). When set, the test pulls a public retail-sales fixture from the AutoGluon S3
+# bucket; otherwise it uses the small CI csv fixture.
+def _load_retail_sales_with_covariates() -> tuple[pd.DataFrame, pd.DataFrame, list[str]]:
+    """Fetch the public retail-sales fixture and split it into train + future-covariates frames.
+
+    Returns ``(train_df, known_covariates_df, known_covariates_names)``. ``known_covariates_df`` shares the schema
+    of ``train_df`` minus the target column, so it carries the future values of the known covariates over the
+    forecast horizon.
```
shchur: Nit: Do we really need this comment?

src/autogluon/cloud/backend/sagemaker_backend.py
```diff
@@ -1009,22 +1060,21 @@ def _upload_fit_artifact(
         if isinstance(train_data, str):
             train_data = load_pd.load(train_data)
         self.original_features = [col for col in train_data.columns if col != label]
-        train_data = self._prepare_data(train_data, "train")
-        logger.log(20, "Uploading train data...")
-        train_input = self.sagemaker_session.upload_data(
-            path=train_data, bucket=cloud_bucket, key_prefix=util_key_prefix
-        )
```
shchur, line 1016: Nice simplification!

shchur (approved):
Looks great, thanks @melopeo!

PR #217, opened 2026-05-12
Add a skeleton for the foundation model class

shchur:
Issue #, if available:

- This PR outlines the user-facing API for the new `FoundationModel` class in AutoGluon Cloud. This class allows users to deploy and run inference with pretrained foundation models (Chronos, Mitra, etc.) on AWS.
- How is this different from `sagemaker.jumpstart.JumpStartModel`?
  - DataFrame in / DataFrame out (no manual JSON payload construction)
  - Supports batch prediction and fine-tuning out of the box
  - Model onboarding is much easier for us in the future - any model available in AutoGluon works as soon as we add it to the config
- How is this different from `TimeSeriesCloudPredictor`?
  - No `fit()` required before `deploy()` or `predict()` — foundation models work out of the box
  - Cleaner API: the user does need to provide nested dictionaries of `predictor_init_kwargs` and `predictor_fit_kwargs`
  - The `FoundationModel` class is stateless — no need to worry about the underlying state (attached artifacts, endpoints, etc)

Example usage
```python
# ============================================================
# Time Series — Chronos
# ============================================================
from autogluon.cloud import FoundationModel

model = FoundationModel("chronos-bolt-base", role_arn="arn:aws:iam::123456789:role/SageMakerRole")

# --- Deploy an endpoint ---
endpoint = model.deploy(instance_type="ml.g5.xlarge")
predictions = endpoint.predict(df, prediction_length=24)
endpoint.delete_endpoint()

# --- Batch predict (large dataset, uses SageMaker Training job under the hood) ---
predictions = model.predict(df, prediction_length=24, quantile_levels=[0.1, 0.5, 0.9])

# --- Batch predict (async) ---
job = model.predict(df, prediction_length=24, wait=False)
# ... later ...
predictions = job.result()

# --- Fine-tune, then deploy ---
fine_tuned = model.fit(train_data=df, instance_type="ml.g5.xlarge")
endpoint = fine_tuned.deploy()

# --- Use custom/cached weights (VPC-safe) ---
model = FoundationModel(
    "chronos-bolt-base",
    hyperparameters={"model_path": "s3://my-bucket/cached-weights/"},
)
endpoint = model.deploy()

# --- Override hyperparameters at call time ---
predictions = model.predict(df, prediction_length=24, hyperparameters={"cross_learning": False})

# ============================================================
# Tabular — Mitra / TabICL
# ============================================================
model = FoundationModel("mitra-classification", role_arn="arn:aws:iam::123456789:role/SageMakerRole")

predictions = model.predict(train_data=train_df, test_data=test_df, label="class")
```

To Do:
- [ ] Wiring the user-facing API to the internal backend that actually runs stuff on AWS

### DIFF
--- a/src/autogluon/cloud/model/__init__.py
+++ b/src/autogluon/cloud/model/__init__.py
@@ -1,0 +1,3 @@
+from .foundation_model import FoundationModel, TabularFoundationModel, TimeSeriesFoundationModel
+
+__all__ = ["FoundationModel", "TabularFoundationModel", "TimeSeriesFoundationModel"]
--- a/src/autogluon/cloud/model/foundation_model.py
+++ b/src/autogluon/cloud/model/foundation_model.py
@@ -162,7 +162,7 @@ class TimeSeriesFoundationModel(FoundationModel):
         timestamp_column: str = "timestamp",
         known_covariates: Optional[Union[str, pd.DataFrame]] = None,
         static_features: Optional[Union[str, pd.DataFrame]] = None,
-        prediction_length: int = 64,
+        prediction_length: int = 1,
         quantile_levels: Optional[List[float]] = None,
         output_path: Optional[str] = None,
         instance_type: Optional[str] = None,
--- a/src/autogluon/cloud/model/foundation_model.py
+++ b/src/autogluon/cloud/model/foundation_model.py
@@ -1,0 +1,233 @@
+"""FoundationModel — deploy and predict with pretrained foundation models on AWS."""
+
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import Any, Dict, List, Literal, Optional, Union
+
+import pandas as pd
+
+from autogluon.cloud.endpoint.endpoint import Endpoint
+from autogluon.cloud.job.remote_job import RemoteJob
+
+from .registry import get_model_config
+
+
+class FoundationModel:
+    """
+    Pretrained foundation model inference on AWS.
+
+    Factory: FoundationModel("chronos-bolt-base", ...) returns the appropriate
+    task-specific subclass (TimeSeriesFoundationModel, TabularFoundationModel).
+
+    Examples
+    --------
+    >>> model = FoundationModel("chronos-bolt-base", role_arn="arn:...")
+    >>> endpoint = model.deploy()
+    >>> predictions = endpoint.predict(data)
+    >>> endpoint.delete_endpoint()
+    """
+
+    def __new__(cls, model_id: str, **kwargs) -> "FoundationModel":
+        if cls is not FoundationModel:
+            return super().__new__(cls)
+        config = get_model_config(model_id)
+        task = config["task"]
+        if task == "timeseries":
+            return super().__new__(TimeSeriesFoundationModel)
+        elif task == "tabular":
+            return super().__new__(TabularFoundationModel)
+        raise ValueError(f"Unsupported task: {task}")
+
+    def __init__(
+        self,
+        model_id: str,
+        backend: Literal["sagemaker"] = "sagemaker",
+        role_arn: Optional[str] = None,
+        region: Optional[str] = None,
+        s3_output_path: Optional[str] = None,
+        model_config: Optional[Dict[str, Any]] = None,
+    ):
+        self.model_id = model_id
+        self.role_arn = role_arn
+        self.region = region
+        self.s3_output_path = s3_output_path
+        self._config = get_model_config(model_id)
+        # Merge user overrides on top of registry defaults
+        self.model_config = {**self._config.get("model_config", {}), **(model_config or {})}
+        # TODO: instantiate backend via BackendFactory
+        self._backend_type = backend
+
+    def deploy(
+        self,
+        instance_type: Optional[str] = None,
+        mode: Literal["realtime", "serverless", "async"] = "realtime",
+        endpoint_name: Optional[str] = None,
+        model_artifact_path: Optional[str] = None,
+        model_config: Optional[Dict[str, Any]] = None,
+        wait: bool = True,
+    ) -> Endpoint:
+        """
+        Deploy model to an endpoint.
+
+        Parameters
+        ----------
+        instance_type
+            Instance type for the endpoint.
+            If None, will use the default from the model registry.
+        mode
+            Endpoint type.
+        endpoint_name
+            Custom endpoint name.
+            If None, will auto-generate a unique name.
+        model_artifact_path
+            S3 path to pre-cached model weights (for VPC / fast cold start).
+            If None, weights are downloaded from HuggingFace on cold start.
+        model_config
+            Override default inference config (prediction_length, quantile_levels, etc.)
+        wait
+            Whether to block until the endpoint is ready.
+
+        Returns
+        -------
+        Endpoint
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def predict(self, data: Union[str, pd.DataFrame], wait: bool = True, **kwargs) -> Union[pd.DataFrame, RemoteJob]:
+        """Subclasses override with task-specific signature."""
+        ...
+
+    def fit(
+        self,
+        train_data: Union[str, pd.DataFrame],
+        output_path: Optional[str] = None,
+        instance_type: Optional[str] = None,
+        wait: bool = True,
+        **kwargs,
+    ) -> "FoundationModel":
+        """
+        Fine-tune the model. Returns a new FoundationModel pointing to the fine-tuned artifact.
+
+        Parameters
+        ----------
+        train_data
+            Training data as DataFrame or S3 path.
+        output_path
+            S3 path to store fine-tuned model.
+            If None, will auto-generate under s3_output_path.
+        instance_type
+            Instance type for the training job.
+            If None, will use the default from the model registry.
+        wait
+            If True, block until training completes.
+
+        Returns
+        -------
+        FoundationModel
+            New instance with model_config pointing to the fine-tuned artifact.
+        """
+        raise NotImplementedError
+
+    def cache_model_artifact(self, s3_path: str) -> str:
+        """
+        Pre-cache model weights to S3 for VPC or production use.
+
+        Launches a small job that downloads weights from HuggingFace
+        and writes them to S3, avoiding large local downloads.
+
+        Parameters
+        ----------
+        s3_path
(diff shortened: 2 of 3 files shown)


AnirudhDagar (commented):
Awesome design and I like this API! Just one thing to note, #213 introduces `autogluon.cloud.init()` which persists the user's `role_arn`, `region`, and S3 bucket to `~/.autogluon/cloud.yaml` at first-time setup. If `FoundationModel` falls back to that config when those args aren't passed, usage collapses to:

```python
# one time per account
import autogluon.cloud as agc
agc.init()

# Every session after that
from autogluon.cloud import FoundationModel

model = FoundationModel("chronos-bolt-base")            # no role_arn needed it will automatically be picked
endpoint = model.deploy(instance_type="ml.g5.xlarge")
predictions = endpoint.predict(data=df, prediction_length=24)
```

I can add this, once both the PRs are shipped.

prateekdesai04:
AG-Cloud requires SageMaker as a package, it would be good to assume that customers may have v3 as well. 
v3 has quite some API changes, do these changes fit v3 and v2 both (should be the case ideally)?
Here is also a migration by the Sagemaker team - https://github.com/nargokul/sagemaker-python-sdk/blob/master/migration.md#migration-tool-mcp-server

abdulfatir:
Thanks @shchur! This looks great and would really simplify the user experience with FMs. A few comments:
- Is `FoundationModel` supposed to be a magic class which would automatically infer the model/task type? Maybe we can add it more concrete in some way, e.g., by adding a `task_type` kwarg or subclassing the base `FoundationModel` class? 
- Related to above: how do we plan to handle model-specific hyperparameters (e.g., `cross_learning`, `n_estimators`, etc.)? What exactly would `model_config` map to internally? 
- For the batch prediction mode, we will have SM job start overhead. Is there a way to reduce this by using another AWS service here (Lambda?)?

src/autogluon/cloud/model/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .foundation_model import FoundationModel, TabularFoundationModel, TimeSeriesFoundationModel
```
shchur, line 1: @prateekdesai04 Thanks Prateek, current plan is to use a `backend` inside the `FoundationModel` class to handle the interactions with AWS. We can add a new backend to handle SageMaker V3 SDK later, I created #218 to track this.

src/autogluon/cloud/model/__init__.py
```diff
@@ -0,0 +1,3 @@
...
+
```
shchur, line 2: @AnirudhDagar thanks Anirudh, makes sense. We can update the logic for automatically resolving the `role` and other constants based on the credentials generating during the setup.

I guess this will also touch a few other places in the code base, so probably best to address this after all other PRs are merged.

src/autogluon/cloud/model/foundation_model.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Examples
+    --------
+    >>> model = FoundationModel("chronos-bolt-base", role_arn="arn:...")
+    >>> endpoint = model.deploy()
+    >>> predictions = endpoint.predict(data)
+    >>> endpoint.delete_endpoint()
+    """
+
+    def __new__(cls, model_id: str, **kwargs) -> "FoundationModel":
```
shchur, line 31: Thanks @abdulfatir, 
1. The `FoundationModel` class automatically resolves to `TimeSeriesFoundationModel`, `TabularFoundationModel`, etc based on the task associated with each `model_id`. The subclasses only differ in their API for `predict`: e.g. `TimeSeriesFoundationModel` takes historical data + known covariates as input, and `TabularFoundationModel` takes train & test DFs as input

src/autogluon/cloud/model/registry.py
```diff
@@ -0,0 +1,60 @@
...
+Maps model_id to AG-compatible configuration for deploy / predict.
+"""
+
+from typing import Any, Dict, Literal, TypedDict
+
+
+class FoundationModelConfig(TypedDict):
+    task: Literal["forecasting", "classification", "regression"]
+    model_name: str  # AG model class name (e.g. "Chronos", "Chronos2", "Mitra")
+    model_config: Dict[str, Any]  # passed to the AG model (e.g. {"model_path": "..."})
```
shchur: @abdulfatir  
2. `model_config` = `hyperparameters` provided to the TimeSeriesPredictor/TabularPredictor under the hood. Now that you pointed this out, I think that we should separate the training and inference configs, e.g.
```python
"chronos-2": {
    "task": "forecasting",
    "model_name": "Chronos2",
    "inference_config": {"model_path": "amazon/chronos-2"},
    "training_config": {"model_path": "amazon/chronos-2", "fine_tune": True},
    ...
},
```
The user can provide kwargs to `deploy()`, `predict()`, `fit()` that updates the default inference/train configs stored in the registry.

Do you think we should allow the user to configure these hyperparameters inside the payload (e.g. passing `cross_learning: True` inside the call to `endpoint.predict`), or is it okay if this is fixed at the creation of the endpoint?

src/autogluon/cloud/model/foundation_model.py
```diff
@@ -0,0 +1,291 @@
...
+        if task == "forecasting":
+            return super().__new__(TimeSeriesFoundationModel)
+        elif task in ("classification", "regression"):
+            return super().__new__(TabularFoundationModel)
+        raise ValueError(f"Unsupported task: {task}")
+
+    def __init__(
+        self,
+        model_id: str,
+        backend: Literal["sagemaker"] = "sagemaker",
```
shchur, line 45: @abdulfatir 
3. We can add support for other backends like Lambda in the future - but I would decouple this from the `FoundationModel` API so that the user doesn't need to worry about it

src/autogluon/cloud/model/__init__.py
```diff
@@ -0,0 +1,3 @@
...
+__all__ = ["FoundationModel", "TabularFoundationModel", "TimeSeriesFoundationModel"]
```
melopeo, line 3: Question:

Are TabularFoundationModel and TimeSeriesFoundationModel part of the public API, or internal to the FoundationModel factory? They're in `__all__`, but examples only use FoundationModel(...). For instance, should users be able to write:
```python
def forecast(model: TimeSeriesFoundationModel, df: pd.DataFrame) -> pd.DataFrame:
    return model.predict(df, prediction_length=24)
```

src/autogluon/cloud/model/foundation_model.py
```diff
@@ -0,0 +1,309 @@
...
+    Pretrained foundation model inference on AWS.
+
+    Factory: FoundationModel("chronos-bolt-base", ...) returns the appropriate
+    task-specific subclass (TimeSeriesFoundationModel, TabularFoundationModel).
+
+    Examples
+    --------
+    >>> model = FoundationModel("chronos-bolt-base", role_arn="arn:...", hyperparameters={"model_path": "s3://cached/"})
+    >>> endpoint = model.deploy()
+    >>> predictions = endpoint.predict(data)
```
melopeo, line 27: nit: 
If the following comment is correct then lets update the example provided in the PR description.

In the description of the PR the example is 
`predictions = endpoint.predict(data=df, prediction_length=24)`
Potentially we should rather use
`predictions = endpoint.predict(df, prediction_length=24)`

The reason is that in `endpoint.py` we have `test_data` and not `data` in predict.
```python
class Endpoint(ABC):
    @property
    @abstractmethod
    def endpoint_name(self) -> str:
        """Name of the endpoint"""
        raise NotImplementedError

    @abstractmethod
    def predict(self, test_data: Union[str, pd.DataFrame], **kwargs) -> Union[pd.DataFrame, pd.Series]:
        """
        Predict with the endpoint
        """
        raise NotImplementedError
```

src/autogluon/cloud/model/foundation_model.py
```diff
@@ -0,0 +1,309 @@
...
+        Endpoint
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def predict(self, data: Union[str, pd.DataFrame], wait: bool = True, **kwargs) -> Union[pd.DataFrame, RemoteJob]:
+        """Subclasses override with task-specific signature."""
+        ...
+
+    def fit(
```
melopeo, line 110: Putting fit() on FoundationModel (the base class) implies every foundation model can be fine-tuned. 

In general it makes sense to support fine-tuning of a pretrained model. My question is if we are assuming all FoundationModels under consideration will have a fine-tuning capability. If a foundation model can't be fine-tuned, how are we planning to handle that?

melopeo:
Hey @shchur , thanks for this. It looks very nice.

Question: The example calls job.result(), but RemoteJob doesn't define it. Is the plan to add it in a follow-up?

src/autogluon/cloud/model/__init__.py
shchur: Yes, with the current design the user should be able to directly create a `TimeSeriesFoundationModel` object bypassing the `FoundationModel` base class, e.g.
```python
model = TimeSeriesFoundationModel(model_id="chronos-bolt-base")
predictions = model.predict(...)
```
The `FoundationModel` base class only exists for convenience.

src/autogluon/cloud/model/foundation_model.py
shchur: Thanks, fixed

src/autogluon/cloud/model/foundation_model.py
shchur: Good question, I added a boolean flag to the `ModelConfig` indicating whether fine-tuning is supported.

src/autogluon/cloud/model/foundation_model.py
```diff
@@ -0,0 +1,316 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def predict(self, data: Union[str, pd.DataFrame], wait: bool = True, **kwargs) -> Union[pd.DataFrame, RemoteJob]:
+        """Subclasses override with task-specific signature.
+
+        When wait=False, returns a RemoteJob handle. Use job.get_job_status() to poll
+        and job.get_output_path() to get the S3 path to predictions once complete.
+        """
+        # TODO: consider adding a .result() method to RemoteJob that downloads + parses output
```
shchur, line 107: @melopeo I missed the top-level comment. Here is a temporary solution, I will figure out the best path when implementing the backend.

melopeo (approved):

PR #220, opened 2026-05-18
Implement batch prediction for `TimeSeriesFoundationModel`

shchur:
Issue #, if available:

Description of changes:
- Wire `FoundationModel` to the SageMaker backend, instantiating it via `BackendFactory` in `__init__`
- Implement `TimeSeriesFoundationModel.predict()` which uses the existing fit_predict pattern under the hood (launches a training job that loads the FM, runs fit + predict, returns results)
- Update model registry to store separate default instance types for `predict`, `deploy` and `fit`
- Rename `s3_output_path` → `cloud_output_path` for consistency with the rest of the module; remove `role_arn` and `region` since these are not currently exposed anywhere in the public API
- Streamline FM tests: share a `retail_sales_dataset` fixture, run the new FM test
- Update the training script `src/autogluon/cloud/scripts/sagemaker_scripts/train.py` to automatically infer the `known_covariates_names` if `known_covariates` DF was provided

### DIFF
--- a/src/autogluon/cloud/scripts/sagemaker_scripts/train.py
+++ b/src/autogluon/cloud/scripts/sagemaker_scripts/train.py
@@ -173,6 +173,9 @@ if __name__ == "__main__":
     if args.known_covariates:
         kc_file = get_input_path(args.known_covariates)
         known_covariates_df = load_pd.load(kc_file)
+        if "known_covariates_names" not in predictor_init_args:
+            kc_cols = known_covariates_df.columns.to_list()
+            predictor_init_args["known_covariates_names"] = kc_cols[2:]  # skip id and timestamp columns
 
     predictor = predictor_cls(**predictor_init_args).fit(training_data, tuning_data=tuning_data, **predictor_fit_args)
 
--- a/src/autogluon/cloud/backend/backend.py
+++ b/src/autogluon/cloud/backend/backend.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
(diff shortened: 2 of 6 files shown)


src/autogluon/cloud/scripts/sagemaker_scripts/train.py
```diff
@@ -173,6 +173,9 @@ def prepare_data(data_file, predictor_type, predictor_init_args=None):
     if args.known_covariates:
         kc_file = get_input_path(args.known_covariates)
         known_covariates_df = load_pd.load(kc_file)
+        if "known_covariates_names" not in predictor_init_args:
+            kc_cols = known_covariates_df.columns.to_list()
+            predictor_init_args["known_covariates_names"] = kc_cols[2:]  # skip id and timestamp columns
```
melopeo, line 178: This is nice! A question that I have here is that we never validate the known_covariates in relationship to the input data and hence if there is any error this will appear once the sagemaker task is submitted. This was previously introduced [here](https://github.com/autogluon/autogluon-cloud/blob/master/src/autogluon/cloud/predictor/timeseries_cloud_predictor.py#L441). Perhaps the main challenge for this is that the current PR calls the backend directly and skips this kind of checks. Do you think this could be addressed somehow in this PR?

melopeo: Lets address this in a follow up PR.

melopeo (approved):

PR #223, opened 2026-05-26
Restructure time series fit/predict data flow

shchur:
Issue #, if available: #70

## Motivation

Pre-PR, the time-series cloud predictor smuggled extra information through positional column hacks. Static features were merged into the train DataFrame as trailing columns past the target; the train and serve scripts then peeled them back off using "anything after the target column is static." Test data was reordered so that `id_column` was at position 0 and `timestamp_column` at position 1, and the serve script inferred those columns from position. Three near-duplicate "convert long-format DataFrame to TimeSeriesDataFrame" helpers lived in the train script, the regular serve script, and the FM serve script — each implementing the same target-column-position trick.

This made the wire format brittle (rename your target column and inference breaks silently), forced a serialization workaround (everything had to fit in one CSV/parquet body), and made the FM deploy path painful: FM has no fit step and can't use the column-position trick at all, so it had to ship its own `_build_tsdf` plus reach into `inference_kwargs` for `id_column` / `timestamp_column`.

## Summary

**Train-time.** Generic `data_channels: Dict[str, DataFrame | str | None]` on `SagemakerBackend.fit()`. Each non-`None` entry rides as its own SageMaker channel (`SM_CHANNEL_TRAIN_DATA`, `SM_CHANNEL_TUNING_DATA`, `SM_CHANNEL_KNOWN_COVARIATES`, `SM_CHANNEL_STATIC_FEATURES`). The train script reads each by name and passes it through `prepare_data` → `TimeSeriesDataFrame.from_data_frame(..., static_features_df=...)` — no column-position hack.

**Inference-time.** `AutoGluonSerializationWrapper` gets explicit `static_features` and `known_covariates` fields. `id_column` / `timestamp_column` are persisted into `predictor_metadata.json` next to the saved model at fit time; the regular serve script's `model_fn` reads them once and `transform_fn` passes them as kwargs to the parsers. `predict()` / `predict_real_time()` no longer accept `id_column` / `timestamp_column` kwargs — the only valid value was "what you said at fit," so it was a footgun.

**Batch predict.** `predict()` (SageMaker batch transform) now serializes `test_data` / `static_features` / `known_covariates` into a single `application/x-autogluon` pickle blob and sends it as one transform request (`split_type="None"`). Previously batch transform couldn't carry side-channel inputs because per-line splitting tore them apart.

**Validation.** New `TimeSeriesSagemakerBackend._validate_data_channels` runs at the top of `fit()`. Validates `train_data` / `tuning_data` / `known_covariates` for required columns and the `known_covariates_names` constraint. Both `TimeSeriesCloudPredictor.fit_predict` and `TimeSeriesFoundationModel.predict` get validation for free since both flow through `backend.fit()`.

**Cleanup along the way.**
- Dropped `predictor_metadata` smuggled into SM hyperparameters (and `attach_job` reading it back) — columns are stamped into the model artifact instead.
- Dropped `self.target_column` / `self.id_column` / `self.timestamp_column` predictor attributes.
- Collapsed three TSDF construction helpers into one `prepare_data` in `train.py`.
- Collapsed `predict_after_fit: bool` and ad-hoc `known_covariates` kwargs on backend `fit()` into a single `extra_ag_args: Dict[str, Any]` extension point.
- Removed the `_preprocess_data` static-features merge hack and the column reorder in batch `predict()`.


melopeo (approved):

src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -269,10 +269,18 @@ def fit(
         )
         if backend_kwargs is None:
             backend_kwargs = {}
+        # Extract data inputs out of predictor_fit_args so they can be uploaded as separate SageMaker
+        # channels. predictor_fit_args itself becomes pure predictor.fit() kwargs.
+        predictor_fit_args = dict(predictor_fit_args)
+        data_channels = {
+            "train_data": predictor_fit_args.pop("train_data"),
+            "tuning_data": predictor_fit_args.pop("tuning_data", None),
+        }
```
garbusbeach, line 278: actually this is a breaking changes if someone (like me) uses custom backend :/ Sad that it was introduced unannonced and proper versioning of this lib has been abandoned :( I cannot even pin the proper version in other way than anonymous commit sha.

Before you ask why one needs a custom backend? I'm using custom train script which accepts also additional test data csv, and we're doing the model test and a few small things on the cloud + store additional leaderboard for that - the additional files upload (+ pass as env vars to sagemaker) is not possible to do in regular sagemaker backend.

That requires a custom Predictor also for backend class resolution, but thats a different story :D

shchur: Hi @garbusbeach, thanks for raising this concern. Can you please clarify what exactly breaks on your side and what would be an acceptable change?

I was assuming that the public API exposed in by `autogluon.cloud` were the 3 predictor classes (`TabularCloudPredctor`, `TimeSeriesCloudPredictor`, `MultiModalCloudPredictor`), and the specific implementations under the hood such as the `Backend` ABC were not a part of the public API. Also, since the package version is currently <1.0, breaking changes between minor releases like `0.4` -> `0.5` are acceptable even under semantic versioning https://semver.org/.

Still, since we haven't yet release v0.5.0, I am happy to review your concerns and see if there's anything we can adjust to make the transition less painful for you.

garbusbeach: I understand :) Actually I started treating 0.4.2 as 1.0.0 long time ago, since there were no updates for a long time ;)
The issue is removing the keys from `predictor_fit_args` (`train_data ` and `tuning_data `) before passing it to the backend - previously that was a backend responsibility to sanitize this values, not the predictor.

I would safely assume that I'm not the one in this case - the autogluon itself is a huge library and it's connections with AWS only ensures me that `autogluon-cloud` is a key element in the ecosystem :)

For me the key feature is beeing able to tweak evaluated python script on sagemaker (which is possible via custom train script kwarg) and also to pass additional data / values to it (ie upload + env var path). In current state it's impossible from what I know, but I'd love to be able to do it.

Basically I want to do some teardown / analysis after training on the machine with data / configs passed programistically.

Also if thats possible I'd be glad to see autogluon roadmap and help in development :)

garbusbeach: Also its worth mentioning I'm using the lib version from master branch, not from pypi (last release jan 2025), because it requires `python <3.12` and `pyarrow <11.1`, which is not the case on master.

For my company it's not possible to downgrade so much - also I'm quite sure that autogluon and autogluon-cloud could safely extend the guards to the more recent versions :)

shchur: Thanks for the clarification. Based on my understanding, it should be a really straightforward fix on your side - just adding something like
```python
predictor_fit_args = {**predictor_fit_args, **data_channels}
```
at the start of the `fit()` method for your custom backend will do the trick.

I don't think there is a clean alternative where we can avoid introducing `data_channels`, this is necessary to have feature parity between `TimeSeriesCloudPredictor` and `TimeSeriesPredictor` where we have two additional input channels (`known_covariates` and `static_features`) that are currently not supported by the Cloud version.

garbusbeach: thanks for clarification!

PR #224, opened 2026-05-27
add predictions_path to fit_predict for direct S3 output

melopeo:
Issue #, if available:

Description of changes:

Currently, fit_predict packages predictions inside the training job's output.tar.gz. To retrieve them, the client must download the tarball, extract predictions.csv, and load it. This PR adds a new optional parameter predictions_s3_path that tells the container to upload predictions directly to a user-specified S3 location, skipping the tarball packaging entirely.

When predictions_s3_path is not set, behavior is unchanged (backwards compatible).

```python
from autogluon.cloud import TimeSeriesCloudPredictor

predictor = TimeSeriesCloudPredictor(
    cloud_output_path="s3://my-bucket/my-run",
)

predictions = predictor.fit_predict(
    train_data=train_df,
    predictor_init_args=dict(target="target", prediction_length=12),
    predictor_fit_args=dict(
        hyperparameters={"Chronos": {"model_path": "bolt_small"}},
    ),
    id_column="item_id",
    timestamp_column="timestamp",
    # NEW: predictions land directly at this S3 path
    predictions_s3_path="s3://my-bucket/my-run/predictions.csv",
)

# predictions is a DataFrame returned in-memory as before.
# The CSV also exists at s3://my-bucket/my-run/predictions.csv.
```

### DIFF
--- a/src/autogluon/cloud/backend/sagemaker_backend.py
+++ b/src/autogluon/cloud/backend/sagemaker_backend.py
@@ -341,15 +341,7 @@ class SagemakerBackend(Backend):
         if extra_ag_args:
             ag_args.update(extra_ag_args)
         if ag_args.get("predict_after_fit"):
-            predictions_path = (
-                ag_args.get("predictions_path") or f"{self.cloud_output_path}/{job_name}/predictions.csv"
-            )
-            if not is_s3_url(predictions_path):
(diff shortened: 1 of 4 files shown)


shchur (approved):
Thanks @melopeo!

PR #228, opened 2026-05-28
Replace pickle wire format with JSON envelope in SageMaker serializers

shchur:
Issue #, if available:

Description of changes:
- Switch `application/x-autogluon*` wire format from pickle to JSON for better security - **breaking change**, old endpoints cannot be used with the new AG-Cloud version
- JSON envelope with base64-encoded parquet for DataFrames, plain JSON list for image byte-strings, and a validated `inference_kwargs` dict
- Add `"version": 1` field to enable clear error messages on future format mismatches
- Remove unused classes `ParquetSerializer` and `JsonLinesSerializer`

### DIFF
--- a/src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
+++ b/src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
@@ -67,11 +67,15 @@ def transform_fn(model, request_body, input_content_type, output_content_type="a
 
     elif input_content_type == "application/x-autogluon-parquet":
         payload = json.loads(request_body)
+        if payload.get("version") != 1:
+            raise ValueError(f"Unsupported x-autogluon payload version: {payload.get('version')}. Expected 1.")
         data = pd.read_parquet(BytesIO(base64.b64decode(payload["data"])))
         inference_kwargs = payload.get("inference_kwargs") or {}
 
     elif input_content_type == "application/x-autogluon-npy":
         payload = json.loads(request_body)
+        if payload.get("version") != 1:
+            raise ValueError(f"Unsupported x-autogluon payload version: {payload.get('version')}. Expected 1.")
         image_bytearrays = [base64.b85decode(_bytes) for _bytes in payload["data"]]
         inference_kwargs = payload.get("inference_kwargs") or {}
 
--- a/src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
+++ b/src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
@@ -1,8 +1,8 @@
 # flake8: noqa
 import base64
 import copy
+import json
 import os
-import pickle
 from io import BytesIO, StringIO
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
```diff
@@ -1,8 +1,8 @@
 # flake8: noqa
 import base64
 import copy
+import json
 import os
-import pickle
```
mbohlkeschneider, line 5: Just out of curiosity: Are we removing pickle completely now?

src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
```diff
@@ -66,24 +66,18 @@ def transform_fn(model, request_body, input_content_type, output_content_type="a
...
 
     elif input_content_type == "application/x-autogluon-parquet":
-        buf = bytes(request_body)
-        payload = pickle.loads(buf)
-        data = pd.read_parquet(BytesIO(payload["data"]))
-        inference_kwargs = payload["inference_kwargs"]
-        if inference_kwargs is None:
-            inference_kwargs = {}
+        payload = json.loads(request_body)
+        if payload.get("version") != 1:
```
mbohlkeschneider, line 70: I don't understand how this versioning works? What is being versioned here?

src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
```diff
@@ -66,24 +66,18 @@ def transform_fn(model, request_body, input_content_type, output_content_type="a
...
-        payload = pickle.loads(buf)
-        data = np.load(BytesIO(payload["data"]), allow_pickle=True)
-        image_bytearrays = []
-        for _bytes in data:
-            im_bytes = base64.b85decode(_bytes)
-            image_bytearrays.append(im_bytes)
-        inference_kwargs = payload["inference_kwargs"]
-        if inference_kwargs is None:
-            inference_kwargs = {}
+        payload = json.loads(request_body)
```
mbohlkeschneider, line 76: Nit - this is probably AI putting in a lot of these validation clauses. I would like to have this in a subfunction that wraps this. Would just make the code a little less cluttered. But not for this PR.

mbohlkeschneider (approved):

src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
shchur: Only for payloads that get sent to the endpoints. We still use it to upload predictor kwargs during fit(), which allows, e.g., to provide custom metrics to the predictor.

src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
shchur: This is the API version. In case we change the structure of the `x-autogluon` payload in the future and use a new AG-Cloud version to invoke an old endpoint, the endpoint will produce a readable error message (below), instead of a cryptic decoding error.

src/autogluon/cloud/scripts/sagemaker_scripts/multimodal_serve.py
shchur: We currently repeat this parsing logic in each script because the serving scripts run on hosts that don't have `autogluon.cloud` installed. This means we have two options:
1. Repeat the loading logic in each serving script
2. Modify the SageMaker model builder to upload the additional `utils.py` file that contains the shared helper functions, alongside the serving script

For now, I think the first option is simpler and less likely to break something unintentionally, but we can revisit this later

src/autogluon/cloud/utils/serializers.py
AnirudhDagar: this is not true anymore, right? maybe update the docstring needs updatng. it's now JSON list of base85 strings, not .npy.

AnirudhDagar (approved):
Looks good feel free to merge, just one minor comment :)

src/autogluon/cloud/utils/serializers.py
shchur: Thanks, good catch

PR #233, opened 2026-05-28
Use logger for setup with default level INFO

AnirudhDagar:
Description of changes:
Replaced prints with INFO level logger.

### DIFF
--- a/src/autogluon/cloud/cli.py
+++ b/src/autogluon/cloud/cli.py
@@ -7,6 +7,7 @@ Rich-based prompts. The Python API is the source of truth for behavior.
 
 from __future__ import annotations
 
+import logging
 from typing import Optional
 
 import boto3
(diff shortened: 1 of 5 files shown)


src/autogluon/cloud/cli.py
```diff
@@ -98,14 +99,20 @@ def bootstrap(
...
+    try:
+        with _console.status(f"Deploying stack '{effective_stack}'...", spinner="dots"):
+            _abort_on_error(
+                _bootstrap,
+                backend=backend,
+                stack_name=effective_stack,
+                session=session,
+            )
+    finally:
+        cloud_logger.setLevel(prior_level)
```
shchur: Minor: maybe move this to a helper on top? smth like
```python
from contextlib import contextmanager

@contextmanager
def _quiet_logger(name: str, level: int = logging.WARNING):
    """Temporarily raise a logger's level so its INFO output doesn't fight the rich spinner."""
    logger = logging.getLogger(name)
    prior = lg.level
    lg.setLevel(level)
    try:
        yield
    finally:
        logger.setLevel(prior)
```
and then the caller becomes simply
```python
with _quiet_logger("autogluon.cloud"), _console.status(f"Deploying stack '{effective_stack}'...", spinner="dots"):
    _abort_on_error(_bootstrap, backend=backend, stack_name=effective_stack, session=session)
```

shchur (approved):
Thanks! One optional suggestion only

src/autogluon/cloud/cli.py
AnirudhDagar: yeah this looks cleaner, updating

